PR #8, opened 2018-11-19
[DPPA-15] Add MibiImage, tiff, and util

jay-ionpath:
Adds MibiImage class as well as tiff and util modules for reading and writing MIBI TIFF files


setup.py
```diff
@@ -16,5 +16,5 @@
           'pip==18.0',
           'requests==2.14.2'
       ],
-      packages=['mibitracker']
+      packages=['mibitracker', 'mibitof']
```
murataksoy: Does this mean that I can import the package mibitof from here ? If so, what if I want to import both this mibitof and the other mibitof in the mibitof repository?

mibitof/tests/test_tiff.py
```diff
@@ -0,0 +1,206 @@
+"""Tests for mibitof.tiff
```
murataksoy: Did you make any changes in these files or did you just copy them over? (would be easier to review if it is the latter)

mibitof/tests/test_util.py
```diff
@@ -0,0 +1,59 @@
+"""Tests for mibiscope.util
```
murataksoy: Not mibiscope.util I guess

setup.py
jay-ionpath: Yes, that is the point. I'm not sure if there will be a naming conflict. I guess we could come up with a different name for this module (mibiprocessing? mibidata?) and then keep mibitof intact.

mibitof/tests/test_tiff.py
jay-ionpath: They are just copy and pasted over.

mibitof/tests/test_util.py
jay-ionpath: Copy and paste for ya

setup.py
murataksoy: I guess that would be better. Another question I have (just curious): when we add ```git+git://github.com/ionpath/mibitracker-client@master``` to environment.yml, does it automatically install both of these modules? I'm just trying to understand how this works.

jay-ionpath: Yes, when pip clones it from git, it runs setup.py which builds the two modules (as well as any others we add in the future). And then it's added it to the path (or some sort of other magic happens) and can just be called as `from mibitof import mibi_image as mi`, etc.

murataksoy: OK, I see that the other mibitof has ```packages=['mibitof']``` statement, so there is a high chance that there will be a naming conflict (maybe you should try it out). Maybe naming as you suggested could be useful and safer.

jay-ionpath: Turns out, yes, there was a naming conflict and so I switched this from `mibitof` to `mibidata`.

.circleci/config.yml
```diff
@@ -28,8 +28,8 @@ jobs:
       - run:
           name: Run tests
           command: |
-            ~/miniconda/envs/mibitracker-client-dev/bin/coverage run -m py.test mibitracker
-            ~/miniconda/envs/mibitracker-client-dev/bin/coverage html -i --omit=mibitracker/tests/*
+            ~/miniconda/envs/mibitracker-client-dev/bin/coverage run -m py.test mibitracker mibidata
+            ~/miniconda/envs/mibitracker-client-dev/bin/coverage html -i --omit=*/tests/*
```
murataksoy: Did you also look at the pylint commands below to take mibidata into account?

mibidata/mibi_image.py
```diff
@@ -0,0 +1,462 @@
...
+            pass
+        try:
+            return [self._mass_index[i] for i in channels]
+        except (KeyError, AttributeError):
+            pass
+        try:
+            return [self._target_index[i] for i in channels]
+        except (KeyError, AttributeError):
+            raise KeyError(f'Subset of channels, targets or massses not found '
+                           'matching {channels}')
```
rfinck: This line is missing an `f''`

mibidata/mibi_image.py
```diff
@@ -0,0 +1,462 @@
+"""Base module for working with MIBI images.
+
+Copyright (C) 2018 Ionpath, Inc.  All rights reserved."""
+
+import datetime
+import os
```
rfinck: Unused import

.circleci/config.yml
```diff
@@ -50,7 +52,7 @@ jobs:
               git checkout gh-pages || git checkout --orphan gh-pages
               git rm -rf .
               cd ..
-              ~/miniconda/envs/mibitracker-client-dev/bin/sphinx-apidoc -f -o docs/source ./mibitracker ./mibitracker/{**,}/tests
+              ~/miniconda/envs/mibitracker-client-dev/bin/sphinx-apidoc -f -o docs/source ./ */tests ./setup.py
```
rfinck: Because the previous one worked for me on my osx laptop (and didn't have the issue you saw on your machine), I checked this version using the circleci local cli and it worked fine.

mibidata/mibi_image.py
rfinck: Nevermind, it's used in a format string and my editor hasn't figured that out yet.

PR #9, opened 2018-11-30
Support for images with tuples channels when saving PNGs

jay-ionpath:
Fix for export_pngs when the image channels are a tuple of (mass, label) instead of either mass or label.

### DIFF
--- a/mibidata/mibi_image.py
+++ b/mibidata/mibi_image.py
@@ -194,8 +194,12 @@ class MibiImage(object):
             self.targets = targets
             self._mass_index = dict(zip(masses, range(length)))
             self._target_index = dict(zip(targets, range(length)))
-        else:
+        elif all(isinstance(c, str) for c in channels):
             self.masses = self.targets = None
+        else:
+            raise ValueError(
+                'Channels must be a list of tuples of (int, str) or a '
+                'list of str')
 
     def __eq__(self, other):
         """Checks for equality between MibiImage instances.
--- a/mibidata/mibi_image.py
+++ b/mibidata/mibi_image.py
@@ -462,5 +462,5 @@ class MibiImage(object):
             im = converter(data[:, :, i])
             png_name = label[1] if isinstance(label, tuple) else label
             with warnings.catch_warnings():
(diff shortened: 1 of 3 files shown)


mibidata/mibi_image.py
```diff
@@ -460,4 +461,6 @@ def export_pngs(self, path, size=None):
         for i, label in enumerate(self.channels):
             im = converter(data[:, :, i])
             png_name = label[1] if isinstance(label, tuple) else label
-            skio.imsave(f'{os.path.join(path, png_name)}.png', im)
+            with warnings.catch_warnings():
```
rfinck: If we wanted to be extra careful, we could check specifically for the low-res warning and ignore only in that case.

jay-ionpath: Yeah, that's probably a good idea.

mibidata/mibi_image.py
```diff
@@ -459,4 +460,7 @@ def export_pngs(self, path, size=None):
             raise TypeError('Unsupported dtype: %s' % data.dtype.type)
         for i, label in enumerate(self.channels):
             im = converter(data[:, :, i])
-            skio.imsave(f'{os.path.join(path, label)}.png', im)
+            png_name = label[1] if isinstance(label, tuple) else label
```
rfinck: Can a label be a tuple of length 1? Or does the setter enforce that it's either a string or length 2?

jay-ionpath: Looks like `_set_channels` checks to ensure each channel is a tuple of length 2.

https://github.com/ionpath/mibitracker-client/blob/71fa9cab7866038510021356b4bd3e9bbde9e4c0/mibidata/mibi_image.py#L185

rfinck: This is a bit of a contrived example, but with that you could set the channels to be 1-tuples:
```
im.channels = [('a',), ('b',), ('c',)]
```
This is an issue with the current channel-setter, your changes *should* have been fine, but maybe we want to enforce that a channel is a 2-tuple or a string here so that we can't have this incompatibility.

PR #10, opened 2018-12-01
Add MibiImage Tutorial Notebook

jay-ionpath:
Tutorial notebook for using the recently added MibiImage class, as well as using tiff.read() and tiff.write()


jay-ionpath:
Already noticed a typo:

`provides conenience methods` (MibiImage Slicing section)

environment_dev.yml
```diff
@@ -4,7 +4,10 @@ channels:
 - defaults
 dependencies:
 - coverage=4.2=py36_0
+- jupyter=1.0.0=py36_7
```
jay-ionpath: We specify that python=3.6.3, should this be =py36_7 or =py36_3?

rfinck: Good question. I copied this from the jupyter that got installed while using python3.6 and only specifying 1.0.0. Let's just take that off so whichever one can work.

PR #13, opened 2019-01-14
Add Search Runs Helper

jay-ionpath:


mibitracker/request_helpers.py
```diff
@@ -160,6 +160,30 @@ def download_file(self, path):
...
+            zero or one.
+        """
+
+        payload = {'name': run_name}
+
+        if run_label:
+            payload['label'] = run_label
+
+        return self.session.get(
+            '{}{}'.format(self.url, self._prepare_route('/runs/')),
```
rfinck: I don't think you need `prepare_route` since you're specifying it here and thus can guarantee to get it right.

jay-ionpath: Good point, I've removed it

PR #15, opened 2019-02-10
DPPA-72 Add script for combining single-plex non-MIBI TIFFs

rfinck:
Also updating copyrights to 2019.


mibidata/combine_tiffs.py
```diff
@@ -0,0 +1,204 @@
...
+if __name__ == '__main__':
+
+    description = ('Generates a single multiplexed MIBItiff from a folder '
+                   'containing individual single-channel TIFFs without MIBI '
+                   'metadata. Note that if the single-channel TIFFs already '
+                   'have the MIBItiff metadata, then a much simpler way to '
+                   'merge them is to use combine_tiffs.merge_mibitiffs '
+                   'instead.')
+
+    parser = argparse.ArgumentParser(description=description)
```
jaytarolli: If I only specify the required args and try to open with BioFormats in ImageJ, I get a NullPointerException (stack below). Adding in instrument, tissue, and aperture fixes this so one or all of these may need some default values if not specified.

(Fiji Is Just) ImageJ 2.0.0-rc-65/1.51w; Java 1.8.0_66 [64-bit]; Windows 10 10.0; 108MB of 12156MB (<1%)
 
java.lang.NullPointerException
at loci.formats.in.IonpathMIBITiffReader.formatMetadata(IonpathMIBITiffReader.java:233)
at loci.formats.in.IonpathMIBITiffReader.initMetadataStore(IonpathMIBITiffReader.java:212)
at loci.formats.in.BaseTiffReader.initMetadata(BaseTiffReader.java:98)
at loci.formats.in.BaseTiffReader.initFile(BaseTiffReader.java:584)
at loci.formats.FormatReader.setId(FormatReader.java:1397)
at loci.plugins.in.ImportProcess.initializeFile(ImportProcess.java:499)
at loci.plugins.in.ImportProcess.execute(ImportProcess.java:142)
at loci.plugins.in.Importer.showDialogs(Importer.java:140)
at loci.plugins.in.Importer.run(Importer.java:76)
at loci.plugins.LociImporter.run(LociImporter.java:78)
at ij.IJ.runUserPlugIn(IJ.java:221)
at ij.IJ.runPlugIn(IJ.java:185)
at ij.Executer.runCommand(Executer.java:137)
at ij.Executer.run(Executer.java:66)
at java.lang.Thread.run(Thread.java:745)

jaytarolli:
My running tally of things that were copied from mibitof into here and should be removed in a followup PR to that repo:
- combine_tiffs.py
- panel.py
- msdf.pyx: parse_xml()
- tests/test_panel.py
- tests/test_msdf.py: test_parse_xml()

In addition, combine_tiffs.spec can also be removed.

mibidata/combine_tiffs.py
rfinck: That is an excellent observation -- it looks like instrument is assumed to be present:
https://github.com/openmicroscopy/bioformats/blob/69499414f5502dbe24126f5770916ebe9d1086bc/components/formats-gpl/src/loci/formats/in/IonpathMIBITiffReader.java#L213

I have added a default value for that to the tiff writer so it's there not only when combining TIFFs but in general.

rfinck: Also, we should add a test using bioformats. Adding a ticket for that now.

mibidata/combine_tiffs.py
```diff
@@ -0,0 +1,204 @@
...
+    parser.add_argument(
+        'folder',
+        help='Folder containing single-channel TIFF files with a .tif or '
+             '.tiff extension.',
+    )
+    parser.add_argument(
+        'run_xml', help='Path to a run XML file.'
+    )
+    parser.add_argument(
+        'point', help='The point number in the run, such as Point1 or Point2.'
```
jaytarolli: A potential pain point for existing users (as it was for me earlier today) was that I was unsure what to use here, so I opened the `sample_parameters.csv` file we packaged with the original version (as I did for other required/optional args). That specified the `point_name` which was `1245_Bottom_R1C4_Tonsil` in that example, which of course fails here because an int is parsed to index the FOV. Not saying this needs to be changed here, but could be something that comes up.

rfinck: Good point (pun intended). Is there a change to the help text you'd suggest?

rfinck: How about something like "... such as Point1 or Point2. This should match the name of the folder created for the raw data."

jaytarolli: Works for me. That should be clear enough for existing users.

rfinck: Cool, I did that plus updated the error message. You approval is requested again.

PR #16, opened 2019-02-10
DPPA-91 Simplify to one conda env for dev, and test setup.py with tox.

rfinck:
There were three different places in here that attempted to maintain a set of python dependencies, which was at least one too many. Also, only one was being tested. So I've removed the non-dev conda environment, renamed the dev one for simplicity, and added tox to test what gets installed via setup.py since we are using that as a depending in other projects.


rfinck:
This is going to cause conflicts with #15 but I think it makes to merge this first because it'll then make sure both our environments work.

PR #21, opened 2019-04-14
DPPA-111 Part 2: Update to turn off paging for most helper functions. 

rfinck:
In the next mibitracker release, pagination will be the default behavior for GET requests that list results, but it can be turned off with `?pagination=no`, which is used here because these are requests that already apply fairly stringent filters.

With the new pagination behavior, results are returned in a dictionary with key `results` that contains an array similar to the old results, except now it's paginated, so also a `count`, `next`, and `previous` keys are returned. The `offset` query parameter can be used to get another page.
```
{
    "count": 7,
    "next": null,
    "previous": null,
    "results": [
        {
            "id": 7,
            "description": "",
            "name": "20180121_1250_1251",
    ...
}

### DIFF
--- a/mibitracker/request_helpers.py
+++ b/mibitracker/request_helpers.py
@@ -200,12 +200,14 @@ class MibiRequests(object):
             A tuple of the response JSON returned by old run and from creating
             its copy, respectively.
         """
-        response = self.get('/runs/?label={}'.format(old_label))
+        # We don't expect there to be many runs with the same label, so it is
+        # safe to turn paging off.
+        response = self.get('/runs/?label={}&paging=no'.format(old_label))
         data = response.json()
         try:
             assert len(data) == 1
         except AssertionError:
-            raise MibiTrackerError('Expected 1 run with label {}, but {} were'
+            raise MibiTrackerError('Expected 1 run with label {}, but {} were '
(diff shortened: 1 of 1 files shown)


jaytarolli:
Is the intent here to get something in so that this doesn't fall out of sync with MIBItracker when the next version is deployed? Because I would be in favor of defaulting to expecting paginated reponses and handling those in a user friendly way. But that's probably not something that would take a chunk of work.

rfinck:
Yes, part of the intent is to get ahead of any upcoming mibitracker changes so this doesn't break, and it's nice to be able to do it in a way that we do have to deploy at exactly the same time but can do this first and have it ready and waiting. 

As far as using pagination in the responses here, I'm open to a follow-up where we do that, although for requests that we can expect will only return few items because we are already applying stringent filters, having pagination off _may_ be the most convenient thing. I would definitely be in favor of using pagination for any additional helper functions that don't apply such stringent filters and thus could return many items.

rfinck:
I tested by running the routine I moved in Part 1 of this to the mibitracker repo, which calls all the helper functions here against a local mibitracker instance. What is key is that I did that with TRKR-4 checked out, and then again with master checked out, to confirm that it is compatible with both the current and future API.

PR #22, opened 2019-04-17
DPPA-179 Script to split scan into pseudo-depths.

rfinck:


mibidata/tests/test_pseudodepths.py
```diff
@@ -0,0 +1,149 @@
...
+        cls.header = struct.pack(pseudodepths.HEADER_FORMAT, 1, 8, NUM_PIXELS)
+        with open(fn, 'wb') as infile:
+            infile.write(cls.header)
+            end_sat = pseudodepths.HEADER_SIZE + \
+                NUM_PIXELS * pseudodepths.SAT_ENTRY_SIZE
+            infile.seek(end_sat)
+            for i, pixel in enumerate(DATA):
+                sat[i, 0] = infile.tell()
+                sat[i, 1] = len(pixel)
+                for loc in pixel:
```
murataksoy: loc -> timestamp maybe?

rfinck:
@murataksoy Yes, it's the same except I updated the tests to compare the raw ion counts rather than the images to liberate it from dependencies.

mibidata/tests/test_pseudodepths.py
```diff
@@ -0,0 +1,149 @@
...
+                    count = int(loc > 0)
+                    infile.write(
+                        struct.pack(pseudodepths.DATA_FORMAT, loc, count))
+            infile.seek(pseudodepths.HEADER_SIZE)
+            for loc, val in sat:
```
murataksoy: Same here: loc, val -> offset, length

mibidata/tests/test_pseudodepths.py
```diff
@@ -0,0 +1,149 @@
...
+
+        self.assertEqual(depth0_header, self.header)
+        self.assertEqual(depth1_header, self.header)
+        self.assertEqual(depth0_sat, self._pack_sat(DEPTH0))
+        self.assertEqual(depth1_sat, self._pack_sat(DEPTH1))
+        self.assertEqual(depth0_data, self._pack_data(DEPTH0))
+        self.assertEqual(depth1_data, self._pack_data(DEPTH1))
+
+    def test_split_into_three(self):
+        with self.assertRaises(ValueError):
```
murataksoy: Some explanation here would be great, such as: This should throw an error because the number of cycles is not divisible by the number of desired splits.

murataksoy:
> @murataksoy Yes, it's the same except I updated the tests to compare the raw ion counts rather than the images to liberate it from dependencies.

Thanks, I just saw your other comment so I looked over the code one more time and added a few minor comments. My decision to approve prevails.

rfinck:
@murataksoy Thanks for the comments, I updated the naming/comments as you suggested and remove one extra print statement. Thus, I again seek your approval.

PR #23, opened 2019-04-18
DPPA-209 Multi (>3) color composites from MibiImages 

### LINKED ISSUE #20: Flipped titles in MibiImage_Tutorial.ipynb, 1st quad-plot
See attached screenshot
Screenshot 2019-03-15 19 50 13

rfinck:
This also includes an update to MibiImage_Tutorial and fixes #20.

Long story short, you can now create an n-color composite (choosing n from our list) with
```
image = MibiImage(...)
composite = mibidata.color.composite(
    image,
    {
        'Green': 'CD3',
        'Cyan': 'CD4',
        'Yellow': 'CD8',
        'Magenta': 'CD68'
    })
### composite is an NxMx3 array that will display with those colors

### CODE REVIEW

File: mibidata/color.py [hunk 1]
```diff
@@ -0,0 +1,149 @@
...
+        gamma: The value with which to scale the image data. Defaults to 1/3.
+            If no gamma correction is desired, set to 1.
+
+    Returns:
+        An NxMx3 uint8 array of an RGB image.
+    """
+    data_map = {}
+    for key, val in color_map.items():
+        data_map[key] = np.power(  # pylint: disable=assignment-from-no-return
+            image[val] / np.maximum(np.max(image[val]), MIN_COUNTS_FOR_SCALING),
```
jaytarolli: I'm not sure how I feel about using the MIN_COUNTS_FOR_SCALING strategy here. That is an opinionated thing we do to ensure images aren't scaled weirdly, but that isn't something a user would necessarily expect this to do. I see a few options here:
- Don't scale the image data other than to put it into the range of 0-1 and require the user to scale on his/her own.
- Add an explanation to the docstring that this is being done so that users aren't completely caught off guard.
- Add a parameter to the function for the value and set the default to MIN_COUNTS_FOR_SCALING so that a user can use his/her own value.

mibidata/color.py
```diff
@@ -0,0 +1,149 @@
...
+    'Red': 0,
+    'Blue': 4 * np.pi / 3,
+}
+# Defines the minimum number of counts to scale images in
+# `composite`. This intended to prevent images with
+# only a few counts from being scaled to [0, 1].
+MIN_COUNTS_FOR_SCALING = 10
+
+
+def hsl2rgb(hsl):
```
jaytarolli: If you're going to pull in `hsl2rgb` here, you may as well pull in `rgb2hsl`. Better to have both in here and not need one than the other way around.

mibidata/color.py
```diff
@@ -0,0 +1,149 @@
...
+        else:
+            screened = _porter_duff_screen(screened, rgb)
+    return screened
+
+
+def composite(image, color_map, gamma=1/3):
+    """Combines multiple image channels by color into a 3-D array.
+
+    Args:
+        image: A MibiImage/
```
jaytarolli: /?

mibitracker/request_helpers.py
```diff
@@ -484,6 +485,22 @@ def image_id(self, run_name, point_name):
 
         return results[0]['id']
 
+    def get_mibi_image(self, image_id):
```
jaytarolli: +1

mibidata/color.py
rfinck: Sure. I even threw in a couple more for free and added example usages the notebook.

mibidata/color.py
rfinck: off-by-one (in keyboard coordinates)

mibidata/color.py
rfinck: Good point. I went with option 3.

PR #24, opened 2019-04-29
DPPA-228: mibitracker-client function to download single channel

jaytarolli:
Similar to `get_mibi_image` from DPPA-209 , a helper function for downloading any single channel such that those that are not SIMS pngs can be downloaded.

### DIFF
--- a/mibitracker/request_helpers.py
+++ b/mibitracker/request_helpers.py
@@ -501,7 +501,7 @@ class MibiRequests(object):
         tiff_data = self.download_file(tiff_path)
         return tiff.read(tiff_data)
 
-    def get_single_channel(self, image_id, channel_name):
+    def get_channel_data(self, image_id, channel_name):
         """Gets a single channel from MIBItracker as a 2D numpy array.
 
         Args:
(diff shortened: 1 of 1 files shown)


mibitracker/request_helpers.py
```diff
@@ -500,6 +501,26 @@ def get_mibi_image(self, image_id):
         tiff_data = self.download_file(tiff_path)
         return tiff.read(tiff_data)
 
+    def get_single_channel(self, image_id, channel_name):
```
rfinck: what do you think of `get_channel_data`?

mibitracker/request_helpers.py
```diff
@@ -500,6 +501,26 @@ def get_mibi_image(self, image_id):
...
+        Returns:
+            A MxN numpy array of the channel data.
+        """
+        image_info = self.get('images/{}/'.format(image_id)).json()
+
+        if not channel_name in image_info['overlays']:
+            raise MibiTrackerError(
+                'Specified channel name is not present in image')
+        buf = io.BytesIO()
+        buf.write(requests.get(image_info['overlays'][channel_name]).content)
```
rfinck: maybe add a little error handling?
```
res = requests.get(image_info['overlays'][channel_name])
res.raise_for_status()
buf.write(res.content)

mibitracker/request_helpers.py
jaytarolli: Yes.

mibitracker/request_helpers.py
jaytarolli: Good call. Done.

PR #25, opened 2019-05-03
DPPA-18: mibitracker-client tutorial that uses segmented data and makes nearest-neighbor plot

jaytarolli:


jaytarolli:
@rfinck I think you said we should do some work on the final figure, but I kind of like it as is. Maybe figure out a way to show the image of those channels behind it? But I think that would be too busy and confusing, and that as is it gets the point across.

rfinck:
Two comments on the notebook:

1. The x-axis limit or some thresholding is weird for CD45 under the "Classify cells" section
2. In the last plot, I think it's be nice to also show a histogram of distances, since that gets at how one can compare across FOVs/cell types.

jaytarolli:
@rfinck 
1. Yeah, that was weird, but it's fixed now.
2. Added the histogram of the nearest neighbor distances

rfinck:
looks good, my last requests are nitpicks on the histogram.
1. we calculate a distance for each CD4+ T cells, so the y-axis label should reflect this. at its simplest it's "number of CD4 T cells", or if we can be more verbose but ultimately every CD4 cell has a value and it's not symmetric to CD8 cells so that should be clear.
2. i would have written "spatial distribution" not "morphological distribution" because i'm not sure we are using morphological appropriately there. but maybe we are?

jaytarolli:
> we calculate a distance for each CD4+ T cells, so the y-axis label should reflect this. at its simplest it's "number of CD4 T cells", or if we can be more verbose but ultimately every CD4 cell has a value and it's not symmetric to CD8 cells so that should be clear.

I changed the y-axis label and added some description to the text above the code cell

> i would have written "spatial distribution" not "morphological distribution" because i'm not sure we are using morphological appropriately there. but maybe we are?

Who knows what I was thinking? Morphological -> spatial.

PR #28, opened 2019-05-21
DPPA 173 - change error message in channel_inds() in mibi_image.py

stanislaw-ionpath:
When mibi_image is initiated with only target names `channel_inds()` function returns a misleading error when called with (mass, taget) tuple.


mibidata/mibi_image.py
```diff
@@ -218,7 +216,7 @@ def __eq__(self, other):
         return False
 
     def __getitem__(self, channels):
-        """Equivalent to :meth:`~slice_data`."""
+        """Equivalent to :meth:`~slijce_data`."""
```
rfinck: ???

mibidata/mibi_image.py
```diff
@@ -192,8 +192,6 @@ def _set_channels(self, channels, length):
                 raise ValueError('Targets are not all unique.')
             self.masses = masses
             self.targets = targets
-            self._mass_index = dict(zip(masses, range(length)))
-            self._target_index = dict(zip(targets, range(length)))
```
rfinck: Nice! Glad to see this nonsense getting removed.

mibidata/mibi_image.py
```diff
@@ -239,30 +237,35 @@ def channel_inds(self, channels):
...
-                           f'matching {channels}')
+            return [self.targets.index(i) for i in channels]
+        except (ValueError, AttributeError):
+            error_msg = f'Subset of channels, targets or massses not found ' \
+                        f'matching {channels}.'
+            if self.targets is None:
+                error_msg += ' Note: channels were indexed with target names ' \
+                             'only.'
+
+            raise KeyError(error_msg)
```
rfinck: I think it'll be clearer if we switch based on whether targets are present rather than appending. Something like:
```
if self.targets is None:
    error_msg = f'Channels were indexed with targets only (no masses were given), available targets are {channels}'
else:
    error_msg = f'Subset of channels, targets or massses not found matching {channels}/'

File: mibidata/mibi_image.py (hunk 1 above)
Comment (stanislaw-ionpath) [reply mibidata/mibi_image.py]: It was an accident... seems like I do not know what I am typing. Even with both hands.

File: mibidata/mibi_image.py (hunk 3 above)
Comment (stanislaw-ionpath) [reply mibidata/mibi_image.py]: Will do

File: mibidata/mibi_image.py [hunk 4]
```diff
@@ -239,30 +237,36 @@ def channel_inds(self, channels):
...
-            raise KeyError(f'Subset of channels, targets or massses not found '
-                           f'matching {channels}')
+            return [self.targets.index(i) for i in channels]
+        except (ValueError, AttributeError):
+            if self.targets is None:
+                error_msg = f'Channels were indexed with targets only (no ' \
+                    f'masses were given), available targets are {channels}'
+            else:
+                error_msg = f'Subset of channels, targets or massses not ' \
+                    f'found matching {channels}/'
```
rfinck: please remove extra '/' at end of line, then this is good to go.

stanislaw-ionpath: done

PR #31, opened 2019-05-22
Dppa 254 docker test

murataksoy:


rfinck:
@murataksoy do you need the custom docker image for this, or were you just trying that while our env wasnt working and now that the env was switched we can swap it back?

murataksoy:
@rfinck  correct, I was testing stuff. I am closing the PR now and deleting the branch.

PR #37, opened 2019-06-12
Adding new segmentation method: circular sectors.

mapazarr:
This is a generalization of the quadrants method. The user can specify the amount of equal sectors in which each cell is divided. The geometric mean is calculated among all sectors.
This method is much more flexible than the quadrants method and helps to better avoid the apparition of false positives. The following figure (using CD8 vs CD4 channels bi-axial plots) depicts the performance of the new method using 8 sectors per cell (left plot), with respect to the quadrant methods (center plot) and the sum of the counts on each cell (right plot):

image_circ_sec_performance

The figure clearly shows a cleaner sample in the case of the circular sectors method. The missclassification is reduced to 1% using the circular sectors method with 8 sectors. This represents an improvement of more than 5% w.r.t. the quadrants method:
```
n_cells: 3153
n misclassification sum: 2200
n misclassification quad: 194
n misclassification circ sec: 30
misclassification sum: 69.77481763399936%
misclassification quad: 6.152870282270853%
misclassification circ sec: 0.9514747859181731%
Improvement of circ sec with n_sec = 8 versus quad method is 5.2013954963526805%
```

The following figure shows in the top the CD8 counts image with the cell boundaries in red (right) and the geometric mean 9using circular sectors) of CD8 counts within each segmented cell (left); in the bottom, the same is shown using the CD4 channel:

image_circ_sec_CD8_and_CD4

The image below shows the performance of the `circular_sectors` method as compared to the `quadrant` method using `%timeit`. The performance of both methods is the same within errors.

image_circ_sec_vs_quad_timeit


mapazarr:
Ok, I saw the error of circleci: it's a few minor errors of pylint about coding style. I will have a look at them soon and correct the style. Otherwise, the tests have passed.

stanislaw-ionpath:
Manuel, this looks great! Please correct the CircleCI errors and I will review it.

rfinck:
@mapazarr Excited to see this! That CD4/CD8 plot looks great, but can you please also show us a figure comparing the CD4 and CD8 *images* too so that we can see that those cells that get pushed to zero match visually with what we'd expect? What I'm referring to is the equivalent of this:
image

mapazarr:
@rfinck I added the picture for CD8 on the first comment. Would you like the CD4 one too?
@stanislaw-ionpath I corrected the errors for pylint. Can you please review the code?

mibidata/segmentation.py
```diff
@@ -20,11 +20,18 @@ def extract_cell_dataframe(label_image, image=None, mode='total'):
...
-            ``'total'`, the ion counts within each labeled region are summed. If
-            ``quadrant``, the geometric mean of each regions's four quadrants
-            is calculated, which favors regions with even spatial distribution.
-            This mode is ignored if an image is not specified.
+        mode: One of``'total'``, ``'quadrant'`` or ``'circular_sectors'``,
+            defaulting to ``'total'``.
+            If ``'total'`, the ion counts within each labeled region are summed.
+            If ``'quadrant'``, the geometric mean of each regions's four
+            quadrants is calculated, which favors regions with even spatial
+            distribution.
```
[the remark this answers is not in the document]
mapazarr: In principle yes

mapazarr:
@rfinck I added also the image of the CD4 channel, though it's more crowded than the CD8 one and hence, not as easily to read. Please let me know if you have any comments on this PR.

mibidata/segmentation.py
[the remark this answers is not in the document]
rfinck: @mapazarr Can you elaborate? Is `n_sec=4` equivalent to quadrants now? Even if it is, are there performance differences?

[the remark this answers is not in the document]
mapazarr: @rfinck OK, I found a bug in the quadrants method: if the geometric center falls inside a pixel, it is not taken into account. I committed some changes to correct this issue and now the circular sectors method with `n_sec=4` gives exactly the same result as the quadrants method. I haven't compared the performance yet, but I think they are similar, the only difference is the polar coordinate transformation, which shouldn't take too long. I can test it further though if it's an important issue.

stanislaw-ionpath:
@rfinck @mapazarr if the circular_sectors can completely reproduce the quadrant I would remove the latter one. If for some historical reasons we need to keep "quadrant" in extract_cell_dataframe let us use circular_sectors with a division to 4 in the backend.

stanislaw-ionpath:
@mapazarr you can use `%timeit` to compare the performance of quadrant and circular section in jupyter.

mapazarr:
@rfinck @stanislaw-ionpath i checked the performance using `%timeit` (see the image below).

image_circ_sec_vs_quad_timeit

The performance of both methods is very similar, so probably we can remove  the quadrant method and keep it as an option that calls circular_sector with `n_sec = 4` as @stanislaw-ionpath suggested.

mibidata/segmentation.py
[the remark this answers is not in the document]
rfinck: Great, let's remove the quadrant method if it's redundant and has a bug.

mibidata/segmentation.py
```diff
@@ -75,11 +86,59 @@ def _quadrant_mean(inds, image):
     upper_left = vals[(inds[0] <= y_center) & (inds[1] < x_center)].sum(axis=0)
     upper_right = vals[(inds[0] < y_center) & (inds[1] >= x_center)].sum(axis=0)
     lower_left = vals[(inds[0] > y_center) & (inds[1] <= x_center)].sum(axis=0)
-    lower_right = vals[(inds[0] >= y_center) & (inds[1] > x_center)].sum(axis=0)
+    lower_right = vals[((inds[0] >= y_center) & (inds[1] > x_center))
+                       | ((inds[0] == y_center) & (inds[1] == x_center))] \
+                       .sum(axis=0)
     quads = np.stack((upper_left, upper_right, lower_left, lower_right), axis=1)
+    # sanity check: test that all pixels in the cell are used
+    assert_array_equal(vals.sum(axis=0), quads.sum(axis=1))
```
rfinck: I think you can remove this now from the final code now that you've tested it.

mibidata/segmentation.py
```diff
@@ -75,11 +86,59 @@ def _quadrant_mean(inds, image):
...
+    vals = image.data[inds]  # has shape (num_pixels_in_cell, num_channels)
+
+    # convert to polar coordinates: y, x -> phi, r
+    inds_pol = np.zeros_like(inds)
+    inds_pol[1], inds_pol[0] = util.car2pol(inds[1], inds[0],
+                                            x_center, y_center)
+    # sanity check: polar coordinate values
+    assert(inds_pol[1] >= 0).all()
+    assert(inds_pol[0] >= 0).all()
+    assert(inds_pol[0] < 360).all()
```
rfinck: Are these necessary in the final version?

mibidata/segmentation.py
```diff
@@ -3,13 +3,14 @@
...
+from numpy.testing import assert_array_equal
 import pandas as pd
 from scipy import ndimage as ndi
 
-from mibidata import mibi_image as mi
+from mibidata import mibi_image as mi, util
 
 
-def extract_cell_dataframe(label_image, image=None, mode='total'):
+def extract_cell_dataframe(label_image, image=None, mode='total', n_sec=8):
```
rfinck: Descriptive rather than abbreviated parameter names are generally prefered, e.g. `sectors` or even `num_sectors` rather than `n_sec`. Final decision up to you though.

mapazarr: done

mibidata/segmentation.py
mapazarr: done

mibidata/segmentation.py
mapazarr: done

mapazarr:
Ok, I removed the quadrants method, cleaned up the code and I even fixed a pylint error (from a file that I didn't change?!). I think the PR is ready to be merged.

mibidata/util.py
```diff
@@ -110,3 +111,23 @@ def format_for_filename(label):
     for char in set(label).intersection(DELIMITERS):
         label = label.replace(char, '-')
     return label
+
+
+def car2pol(x, y, x_c=0, y_c=0):
+    """Convert cartesian to polar coordinates w.r.t. a central point.
+    Angle phi is returned in deg in the range [0, 360) deg.
```
rfinck: Is there a reason to do this in degrees and not radians?

mapazarr: I find it more user friendly to use degrees instead of radians, but I can change it if you prefer radians.

mapazarr: done

mibidata/segmentation.py
```diff
@@ -49,35 +57,52 @@ def extract_cell_dataframe(label_image, image=None, mode='total'):
...
-    upper_right = vals[(inds[0] < y_center) & (inds[1] >= x_center)].sum(axis=0)
-    lower_left = vals[(inds[0] > y_center) & (inds[1] <= x_center)].sum(axis=0)
-    lower_right = vals[(inds[0] >= y_center) & (inds[1] > x_center)].sum(axis=0)
-    quads = np.stack((upper_left, upper_right, lower_left, lower_right), axis=1)
-    return np.power(np.product(quads, axis=1), 1 / 4)
+
+    # convert to polar coordinates: y, x -> phi, r
+    inds_pol = np.zeros_like(inds, dtype=float) # force float for accuracy
+    inds_pol[1], inds_pol[0] = util.car2pol(inds[1], inds[0],
+                                            x_center, y_center)
```
rfinck: Since you're only using the angle, this can be simplified to
```
_, angles = util.car2pol(inds[1], inds[0], x_center, y_center)

File: mibidata/segmentation.py [hunk 7]
```diff
@@ -49,35 +57,52 @@ def extract_cell_dataframe(label_image, image=None, mode='total'):
...
+
+    # create circular sectors
+    sectors = []
+    ang_step = 2.*np.pi/num_sectors
+    for i in range(num_sectors):
+        new_sector = vals[(inds_pol[0] >= i*ang_step)
+                          & (inds_pol[0] < (i + 1)*ang_step)].sum(axis=0)
+        sectors.append(new_sector)
+    secs = np.stack(sectors, axis=1)
```
rfinck: Here you are essentially binning a histogram with edges of angles, weighted by pixel counts, so would something like the following work?
```
edges = np.linspace(0, 2*np.pi, num_sectors+1)
secs = np.bincount(np.digitize(angles, edges),
                   weights=vals, minlength=num_sectors)
```

mibidata/segmentation.py
mapazarr: Ok, I'm using now:
```python
phi = util.car2pol(inds[1], inds[0], x_center, y_center)[1]
```

mibidata/segmentation.py
mapazarr: Ok, I tried your method, with a few modifications. The major one being: a loop for each channel in the image. Unfortunately ```np.bincount``` only works in 1 dim. Please have a look at the new version and let me know your thoughts.

mapazarr: @rfink: I reverted to the previous implementation because the check for empty sectors is more straightforward in the previous implementation. More details below in a new comment. Please let me know of your thoughts.

mapazarr:
@rfinck @stanislaw-ionpath I added a new test for the method that should cover the possibility of encountering small cells (i.e. with few pixels in the image) or strangely shaped cells (i.e. elongated cells). The idea of the test is simple: what happens if we have an empty sector in the cell?

This new test implements a small cell that leaves many sectors of the circular sector method empty. Basically there are more sectors defined than pixels in the image. By default, the circular sectors method would fill a 0 for these empty sectors. But then, since the method calculates a geometric mean of all sectors, the whole cell would be set to zero just because one (or more) sectors are empty. This is not correct. Please notice that the same was happening with the old quadrant method that we removed.

I now added a check to the circular sectors method: if a sector is empty, instead of filing a zero, a one is filled, which is the neutral element of the multiplication operation involved in the geometric mean. I think this is still not 100% correct, but it's much better than the case before.

The correct implementation would be to discard the empty sector and lower the number of sectors for that particular cell. This is more complicated to implement and probably more computationally demanding. Since the amount of cells affected should be small (most cells have a roundish shape and are imaged using many pixels), it's probably not worth the effort of implementing this dynamic number of sectors.

mapazarr:
I just cleaned up a little bit the commit history of this branch. I squashed some minor commits to have a more compact log history.

@rfink @stanislaw-ionpath: I am done with this branch. Maybe you can perform a final review and let me know if it can be merged or if we still need to implement/review some changes.

mibidata/segmentation.py
rfinck: It's fine to keep your implementation if it works better here due to the dimensionality etc. Thanks for looking into other options.

mibidata/segmentation.py
```diff
@@ -49,35 +57,55 @@ def extract_cell_dataframe(label_image, image=None, mode='total'):
...
+    # create circular sectors
+    sectors = []
+    ang_step = 2.*np.pi/num_sectors
+    for i in range(num_sectors):
+        values = vals[(phi >= i*ang_step) & (phi < (i + 1)*ang_step)]
+         # check if the sector is empty; if so, fill one (neutral element for
+         # the multiplication in the geometric mean); otherwise the whole cell
+         # will be set to zero
+        if not values.size:
+            values = np.ones((1, len(image.channels)))
```
rfinck: I have mixed feelings about this -- it was intentional in the quadrants scheme to send everything to zero if the marker was not present in all quadrants. But I understand that when there are more sectors this can bias more things to zero than we'd like --that's partly what I wanted to see in [this comment](https://github.com/ionpath/mibitracker-client/pull/37#issuecomment-501356369). I wonder if we want to make it an option to use zeros or ones here. Having it default to zero would also make this backwards-compatible with our previous usage.

stanislaw-ionpath:
> squashing/merging you did makes it unclear in the PR which changes are from you as opposed to others that were merged to master in the meantime. If it's not too much trouble for you to remove those extra commits that'd make the history much cleaner.

@rfinck do not blame @mapazarr. It was me merging to master after my last review. I remerged again and all the strange looking changes are gone.

mibidata/segmentation.py
stanislaw-ionpath: @rfinck I support @mapazarr implementation. Values are set to 1 only if the size of the sector is 0. 
However, I think it would be advisable to throw a warning whenever that happens.

mapazarr: @rfink the behavior you describe from the quadrants method is kept in the current implementation: if the marker is not present in a sector it's set to zero, hence the cell is discarded. BUT, if the sector is empty (meaning there are no pixels), there is no information if the marker is present or not. In the old quadrants implementation a zero would be set for the sector, discarding the whole cell, although we don't have enough information for discarding it, since the sector is empty. My implementation sets a one in order to let the sectors that contain pixels decide if the cell is kept or not. Please let me know if you need more clarification or if you want me to change something in the code. Throwing a warning as @stanislaw-ionpath suggests might be a good idea. Let me know if you agree.

mapazarr:
> > squashing/merging you did makes it unclear in the PR which changes are from you as opposed to others that were merged to master in the meantime. If it's not too much trouble for you to remove those extra commits that'd make the history much cleaner.
> 
> @rfinck do not blame @mapazarr. It was me merging to master after my last review. I remerged again and all the strange looking changes are gone.

@rfink I'm sorry that I squashed without asking first. I didn't change anything in the commits that didn't belong to this PR. I had to incorporate them to the branch because they were merged to master before this PR. The fix from @stanislaw-ionpath should have solved the issue. If not, please let me know and I'll have a look into it.

mibidata/segmentation.py
rfinck: Thanks for clarifying my misunderstanding. This sounds fine as it is now.

PR #39, opened 2019-06-21
DPPA-294: Add merge_masses as utility function to mibitracker-client

NoahNewgren:


NoahNewgren:
Actually, this function is already in mibitracker-client, but is a private function (mibidata.panels._merge_masses). I think this should 1) be rewritten as a public function with 2) proper documentation and maybe some testing.

mibidata/panels.py
```diff
@@ -56,10 +56,20 @@ def read_csv(path):
...
-        return _merge_masses(combined)
+        return merge_masses(combined)
 
 
-def _merge_masses(df):
+def merge_masses(df):
+    """Merges 'Target' cells of a dataframe with the same 'Mass' value.
+
+    Args:
+        dataframe: A dataframe of the CSV file containing columns 'Mass' and
```
jaytarolli: The first part before the ':' should match the name of the parameter in the function definition, i.e. `df` in this case.

mibidata/panels.py
```diff
@@ -56,10 +56,20 @@ def read_csv(path):
...
+        'Target'.
+
+    Returns:
+        A dataframe containing columns 'Mass' and 'Target' with merged targets
```
jaytarolli: A DataFrame containing...

mibidata/panels.py
```diff
@@ -57,10 +57,20 @@ def read_csv(path):
...
+        return merge_masses(combined)
 
 
-def _merge_masses(df):
+def merge_masses(df):
+    """Merges 'Target' cells of a dataframe with the same 'Mass' value.
+
+    Args:
+        df: A DataFrame of the panel containing columns 'Mass' and
+        'Target'.
```
jaytarolli: Indent this line by one tab.

mibidata/panels.py
```diff
@@ -57,10 +57,20 @@ def read_csv(path):
         # Combine and convert 'Mass' column to int
         combined = pd.concat(df, ignore_index=True)
         combined['Mass'] = combined['Mass'].astype(np.int64)
-        return _merge_masses(combined)
+        return merge_masses(combined)
 
 
-def _merge_masses(df):
+def merge_masses(df):
+    """Merges 'Target' cells of a dataframe with the same 'Mass' value.
```
jaytarolli: ...a DataFrame with...

mibidata/panels.py
```diff
@@ -57,10 +57,20 @@ def read_csv(path):
...
+
```
jaytarolli: The sentence above is a good one line summary of the function. For docstrings in our code, we follow the convention that the summary is a single line (as it is here) and when appropriate, additional details or a more long-winded description can be added after a blank line. For this docstring, could you add "This function merges multiple targets that are conjugated to the same mass tag such that the returned DataFrame contains only unique masses. Target names are combined using the conventions of `util.natural_sort()`." as additional details?

murataksoy: Small addition: in the docstrings, you can also reference other functions, such as the `util.natural_sort()` here such that, when the docs are automatically created after your merge, one can click on this function and immediately go to the doc of that function. It pretty much creates a hyperlink.

Though I don't know how to do this to some function in another python file as in this case. I leave it up to you to figure it out ;-)

jaytarolli: For fear of you wasting time looking up how to do it, @murataksoy and I just figured out it's :func:`function_name` where you replace function_name with the function you are trying to link to.

PR #40, opened 2019-06-24
DPPA-95 image_id() returns incorrect error when no results found

stanislaw-ionpath:
image_id() returns incorrect error when no results found


mibitracker/request_helpers.py
```diff
@@ -475,13 +475,16 @@ def image_id(self, run_name, point_name):
         ).json()
 
         try:
-            assert len(results) == 1
+            len_results = len(results)
```
rfinck: This'll work, although is there a need to keep the TypeError clause or can we go straight to checking the length?

stanislaw-ionpath: @rfinck I don't know what is actually returned by the API call. Where can I check that?

rfinck: You can make a request to our instance and see what you get back? (Also, please edit the name of this PR to have a brief description in addition to the ticket number.)

stanislaw-ionpath: @rfinckit seems we do not need TypeError.

rfinck: no one has ever called me rfinckit before ...

if we don't need the TypeError, please remove it and then we can merge this it.

stanislaw-ionpath: @rfinck done

mibidata/mibi_image.py
```diff
@@ -261,10 +261,11 @@ def channel_inds(self, channels):
         except (ValueError, AttributeError):
             if self.targets is None:
                 error_msg = f'Channels were indexed with targets only (no ' \
-                    f'masses were given), available targets are {channels}'
+                    f'masses were given), available targets are ' \
+                    f'{self._channels}'
```
jaytarolli: Sorry if I missed this, but what is the purpose of changing from `channels` to `self._channels` here?

stanislaw-ionpath: This is a separate bug that I noticed. 

`self._channles` are channels available in the mibiimage, `channels` are channels that the function should return.

jaytarolli: The way it was written made sense to me, so maybe I was misinterpreting it all along. The new error message added here, as you have written it, makes sense. But `Subset of channels, targets or massses not found matching ...` makes sense with the channels that were passes as an argument and not the full list of channels in the image.

PR #41, opened 2019-06-24
DPPA-283: Enforce order of target names when they share the same mass

NoahNewgren:
We have "HLA Class 1" and "Na-K-ATPase" on 176, but they appear twice on the target list because we probably did not enforce their order when combining their names.

### DIFF
--- a/mibidata/panels.py
+++ b/mibidata/panels.py
@@ -62,7 +62,7 @@ def read_csv(path):
 
 def _merge_masses(df):
     conjugates = {}
-    targ_list = []
+    target_list = []
     for conj in df.to_dict(orient='records'):
         mass = conj['Mass']
         target = conj['Target']
@@ -71,7 +71,7 @@ def _merge_masses(df):
         else:
             conjugates[mass] = [target]
     for mass in conjugates:
-        targ_list = conjugates[mass]
-        util.natural_sort(targ_list)
-        conjugates[mass] = ', '.join(targ_list)
+        target_list = conjugates[mass]
+        util.natural_sort(target_list)
+        conjugates[mass] = ', '.join(target_list)
     return pd.DataFrame(list(conjugates.items()), columns=['Mass', 'Target'])
--- a/mibidata/panels.py
+++ b/mibidata/panels.py
@@ -5,6 +5,7 @@ Copyright (C) 2019 Ionpath, Inc.  All rights reserved."""
 import numpy as np
 import pandas as pd
 from pandas.errors import ParserError
+from mibidata import util
 
 
 def read_csv(path):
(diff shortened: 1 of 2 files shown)


mibidata/tests/test_panels.py
```diff
@@ -119,6 +155,19 @@ def test_read_tracker_panel_with_empty_cells(self):
 
         pd.testing.assert_frame_equal(loaded, self.expected_df)
 
+    def test_merge_panels_with_repeated_masses(self):
```
jaytarolli: I should have mentioned this earlier, but these tests should really just encompass the code changes you made, which was to `_merge_masses()`. For example, it's not necessary to test reading the panel CSV, which the other unit tests in here do. Rather, in each unit test you've added, you can create two DataFrame objects, one the input and one the expected value (they don't need to be class level variables unless you wind up reusing say the expected DataFrame across multiple unit tests) and then assert that the return value from `_merge_masses()` matches the expected.

```
df_input = pd.DataFrame({'Mass': (where masses are repeated)..., 'Target': ...})
merged_df = _merge_masses(df_input)

expected_df = pd.DataFrame({'Mass': ..., 'Target': (where targets are properly merged)...})
pd.testing.assert_frame_equal(merged_df, expected_df)
```

In this way, you are testing only the function and leaving the reading function (even though it itself calls `merge_masses()`) to other unit tests.

mibidata/panels.py
```diff
@@ -61,11 +62,16 @@ def read_csv(path):
 
 def _merge_masses(df):
     conjugates = {}
+    targ_list = []
```
jaytarolli: I prefer to avoid using shortened variable names for readability, so could you rename this to `target_list`? Ask @murataksoy what happens when he tries to submit a PR for C++ code with variable names `int aaa = 0`.

mibidata/panels.py
```diff
@@ -61,11 +62,16 @@ def read_csv(path):
...
         target = conj['Target']
         if conjugates.get(mass):
-            conjugates[mass] = f'{conjugates[mass]}, {target}'
+            conjugates[mass].append(target)
         else:
-            conjugates[mass] = target
+            conjugates[mass] = [target]
+    for mass in conjugates:
+        target_list = conjugates[mass]
+        util.natural_sort(target_list)
```
murataksoy: I think natural sort is not really needed here, but it doesn't hurt to use it either.

jaytarolli: It's not necessary for enforcing the same order each time, yes, but it will make ordering consistent with other places in MIBItracker, mibitof, etc.

murataksoy: I meant we don't have targets like Target1, Target2, ...... Target10. The standard function `conjugates[mass].sort()` could also work fine.

jaytarolli:
We have targets like CD3, CD31, etc. I’m sure we’ll reach a point where a panel is filled with “CDn” where n is a 1-3 digit integer.

PR #43, opened 2019-07-03
DPPA-289 Add default print to mibi-image 

stanislaw-ionpath:
__str__() and __repr__() functions that return a string with information from MibiImage.metadata()

### DIFF
--- a/mibidata/mibi_image.py
+++ b/mibidata/mibi_image.py
@@ -220,13 +220,12 @@ class MibiImage():
         return self.slice_data(channels)
 
     def __repr__(self):
-        s = f"{type(self)}\n"
-        for key, val in self.metadata().items():
-            s += f'{key}: {val}\n'
+        s = f'{type(self)}\n'
+        s += '\n'.join(f'{key}: {val}' for key, val in self.metadata().items())
         return s
 
     def __str__(self):
-        s = f"{type(self)} " + "{"
+        s = f'{type(self)} ' + '{'
         s += ', '.join(f'{key}: {val}' for key, val in self.metadata().items())
         s += '}'
         return s
--- a/mibidata/mibi_image.py
+++ b/mibidata/mibi_image.py
@@ -219,6 +219,18 @@ class MibiImage():
         """Equivalent to :meth:`~slice_data`."""
         return self.slice_data(channels)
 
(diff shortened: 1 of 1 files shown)


mibidata/mibi_image.py
```diff
@@ -219,6 +219,18 @@ def __getitem__(self, channels):
         """Equivalent to :meth:`~slice_data`."""
         return self.slice_data(channels)
 
+    def __repr__(self):
+        s = f"{type(self)}\n"
+        for key, val in self.metadata().items():
+            s += f'{key}: {val}\n'
+        return s
+
+    def __str__(self):
```
rfinck: `__str__` will default to be the same as `__repr__`, so you can omit this here.

mibidata/mibi_image.py
```diff
@@ -219,6 +219,18 @@ def __getitem__(self, channels):
         """Equivalent to :meth:`~slice_data`."""
         return self.slice_data(channels)
 
+    def __repr__(self):
+        s = f"{type(self)}\n"
```
rfinck: For consistency please use single-quotes `'`

mibidata/mibi_image.py
```diff
@@ -219,6 +219,18 @@ def __getitem__(self, channels):
...
+        for key, val in self.metadata().items():
+            s += f'{key}: {val}\n'
```
rfinck: For a dict of this length it's not a big deal, but in general it's better to do
```
'\n'.join(f'{key}: {val}' for key, val in im.metadata().items())
```
because `+=` creates a new string at each iteration.

mibidata/mibi_image.py
stanislaw-ionpath: I tried that, but then I get \n in the string which looks ugly.

mibidata/mibi_image.py
stanislaw-ionpath: done

mibidata/mibi_image.py
stanislaw-ionpath: done

mibidata/mibi_image.py
rfinck: Ah I hadn't noticed the distinction between the two. Got it.

PR #45, opened 2019-07-11
DPPA-353 - mibitracker-client `No images found matching` error

stanislaw-ionpath:
I got the following error:`mibitracker.mibitracker_exceptions.MibiTrackerError: No images found matching run 20190710_2201_1 Point2. error`but I can access 20190710_2201_1 Point2 in mibitracker. That this is because the image_id method in request_helper.py is checking not the run__name, but run__label.


stanislaw-ionpath:
> Please update the docstring to indicate that the run label and then the run name will be checked.

done

chris-brent:
@rfinck  Aren't run_name and run_label a composite key? As coded above if I have RunA and RunB and both have an image called Point1 and the run_label LabelA we'll return the:Multiple images match run {run_name} {point_name}.
Error. A better way would be to have an optional run_label parameter that is used in the image search.

stanislaw-ionpath:
@chris-brent you pushed your email and password

chris-brent:
> @chris-brent you pushed your email and password

Well then.....

rfinck:
This function had some parameters that weren't well-named (my fault) but the intended and typical use of it is to fetch a single image id. Since run_label has the uniqueness constraint and thus can be relied upon to get a single run, it should be what is primarily queried, as it was in the previous version of this function. I'm going to make a couple changes to reflect that since this branch has gotten into a bit of a weird state, but next time let's try to do this with comments instead.

stanislaw-ionpath:
@rfinck I made small text edits to make it more straight forward for non-native speakers :)

PR #46, opened 2019-07-14
DPPA-355 Fix reading FOV size from TIFF by not assuming denominator of 1

rfinck:
This issue would only arise with certain FOV sizes, such that pixels / cm fraction had a denominator other than one. It affected the attribute on a MibiImage, but not what was written into any TIFFs.

For example, a 500 micron 1024 x 1024 FOV would not have been affected, because
```
(10000 microns / cm) * (1024 pixels) / (500 microns) = 20480 pixels / cm
```
whereas a 300 micron FOV would have a nonzero denominator when represented as a fraction
```
(10000 microns / cm) * (1024 pixels) / (300 microns) = 34133.333... pixels / cm
```

### DIFF
--- a/mibidata/tiff.py
+++ b/mibidata/tiff.py
@@ -279,11 +279,13 @@ def _page_description(page):
 def _page_metadata(page, description):
     """Parses the page metadata into a dictionary."""
     assert page.tags['resolution_unit'].value == 3
-    assert page.tags['x_resolution'].value[0] ==\
-        page.tags['y_resolution'].value[0],\
-        'x-resolution and y-resolution are not equal'
-    size = (page.tags['image_width'].value /
-            float(page.tags['x_resolution'].value[0]) * 1e4)
+    x_resolution = page.tags['x_resolution'].value[0] / \
+        page.tags['x_resolution'].value[1]
+    y_resolution = page.tags['y_resolution'].value[0] / \
+                   page.tags['y_resolution'].value[1]
+    assert (x_resolution == y_resolution,
+            'x-resolution and y-resolution are not equal')
+    size = page.tags['image_width'].value / x_resolution * 1e4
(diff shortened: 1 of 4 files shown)


mibidata/tiff.py
```diff
@@ -279,11 +279,13 @@ def _page_description(page):
 def _page_metadata(page, description):
     """Parses the page metadata into a dictionary."""
     assert page.tags['resolution_unit'].value == 3
-    assert page.tags['x_resolution'].value[0] ==\
-        page.tags['y_resolution'].value[0],\
+    x_resolution = page.tags['x_resolution'].value[0] / \
+        page.tags['x_resolution'].value[1]
+    y_resolution = page.tags['y_resolution'].value[0] / \
+                   page.tags['y_resolution'].value[1]
```
jaytarolli: jaylint here: The line above this has one tab indent for the line continuation and this one is lined up with the = sign.

rfinck:
@jaytarolli The issue was that the resolution is reported as a _rational_ number, so we must divide the numerator `page.tags['x_resolution'].value[0]` by the denominator `page.tags['x_resolution'].value[1]` to get the correct float. Before we were just grabbing the numerator, but In many cases the denominator is 1 so our tests passed and the values we've been using have mostly made sense.

mibidata/tiff.py
rfinck: fixed

jaytarolli:
@rfinck Got it now, thanks!

mibidata/tiff.py
jaytarolli: Oh, I liked the other way better 😛

yari-ionpath:
late to the game as I was looking through some of the terminology in the file to understand what was being done. The fix makes sense to me too.

PR #47, opened 2019-07-25
MibiRequests can be initialized with JWT for mibitracker authentication

rileysparsons:
`MibiRequests` has a new signature.

It will now accept either a JWT token or an email or password.

### DIFF
--- a/mibitracker/request_helpers.py
+++ b/mibitracker/request_helpers.py
@@ -80,6 +80,7 @@ class MibiRequests():
             self.session.headers.update({
                 'Authorization': 'JWT {}'.format(token)
             })
+            self.session.options(self.url)
         elif email is not None and password is not None:
             self._auth(url, email, password)
         else:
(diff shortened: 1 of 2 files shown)


mibitracker/tests/test_request_helpers.py
```diff
@@ -39,6 +39,36 @@ def setUp(self):
...
+
+    def test_parameter_validation(self):
+        try:
+            request_helpers.MibiRequests(
+                'https://mibitracker-instance.ionpath.com',
+                None,
+                None,
+                None
+            )
+        except ValueError:
```
rfinck: Thanks for adding these tests! I'd typically use the pattern below, but what you did also makes sense and I wasn't familiar with the `fail` method and will be sure to find it useful.
```
    request_helper.MibiRequests(...)
```

mibitracker/request_helpers.py
```diff
@@ -61,13 +62,30 @@ class MibiRequests():
...
         self.url = url.rstrip('/')  # We add this as part of request params
         self.session = StatusCheckedSession()
-        self._auth(url, email, password)
+
+        # Provide either an email and password, or a token
+        # The token will be used in lieu of email and password if provided
+        if token is not None:
+            self.session.headers.update({
+                'Authorization': 'JWT {}'.format(token)
+            })
```
rfinck: This looks great but as you just mentioned it'd be even better if you validated the token here too.

PR #50, opened 2019-08-09
DPPA-390: Panel CSV cannot merge masses when targets all numbers

jaytarolli:


jaytarolli:
I installed mibitacker-client into mibitof from this branch and ran MIBI/O to confirm the panel csv file in the ticket will load without error:
image

PR #58, opened 2019-09-16
DPPA-457: Update copy_run to use new MIBItracker attributes

chris-brent:
Update the copy_run function to use the new fov_size attribute.

### DIFF
--- a/mibitracker/request_helpers.py
+++ b/mibitracker/request_helpers.py
@@ -242,6 +242,8 @@ class MibiRequests():
             'instrument': data['instrument']['id'],  # required
             'slides': ','.join(str(d['id']) for d in data['slide_ids']),
             'label': new_label,
+            'fov_size': data['fov_size'],
+            'aperture': data['aperture'] and data['aperture']['id'],
             'project': data['project'] and data['project']['id'],  # optional
             'description': data['description'],
             'operator': data['operator'],  # Not yet used, but field exists
@@ -288,7 +290,6 @@ class MibiRequests():
                     new_run_label, item['folder']))
             assert len(response.json()) == 1
             new_image = response.json()[0]
-
             # Update the section and tissue of the copied image using
             # info from the original image
             updated_image = {
@@ -297,9 +298,6 @@ class MibiRequests():
                 'project': item['section'] and \
                            item['section']['slide']['project']['id'],
                 'fov_size': item['fov_size'],
-                'aperture': item['aperture'] and item['aperture']['id'],
-                'imaging_preset': item['imaging_preset'],
-                'lens1_voltage': item['lens1_voltage']
             }
             # Add optional updates to copy such as a different frame size due to
             # cropping, or fixed mass calibration, etc.
--- a/mibitracker/request_helpers.py
+++ b/mibitracker/request_helpers.py
@@ -243,7 +243,7 @@ class MibiRequests():
             'slides': ','.join(str(d['id']) for d in data['slide_ids']),
             'aperture': data['aperture']['id'],  # required
             'label': new_label,
-            'magnification': data['magnification'],  # required, FOV size
+            'fov_size': data['fov_size'],  # required, FOV size
             'project': data['project'] and data['project']['id'],  # optional
             'description': data['description'],
             'operator': data['operator'],  # Not yet used, but field exists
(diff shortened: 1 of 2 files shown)


stanislaw-ionpath:
@rachel I run updated this PR and tested the code. PLease have a look

mibitracker/request_helpers.py
```diff
@@ -241,9 +241,9 @@ def copy_run(self, old_label, new_label, **kwargs):
         run_data = {
             'instrument': data['instrument']['id'],  # required
             'slides': ','.join(str(d['id']) for d in data['slide_ids']),
-            'aperture': data['aperture']['id'],  # required
             'label': new_label,
-            'magnification': data['magnification'],  # required, FOV size
+            'fov_size': data['fov_size'],
+            'aperture': data['aperture'] and data['aperture']['id'],
```
rfinck: Aperture also moved to the Image model, we kept it on Run for backwards compatibility but please also specify it at the image level so that if it varies by image that gets captured.

mibitracker/request_helpers.py
```diff
@@ -241,9 +241,9 @@ def copy_run(self, old_label, new_label, **kwargs):
         run_data = {
             'instrument': data['instrument']['id'],  # required
             'slides': ','.join(str(d['id']) for d in data['slide_ids']),
-            'aperture': data['aperture']['id'],  # required
             'label': new_label,
-            'magnification': data['magnification'],  # required, FOV size
+            'fov_size': data['fov_size'],
```
rfinck: `'fov_size'` gets ignored here, here we also retained the ability to post with `'magnification'` for backwards compatibility.

mibitracker/request_helpers.py
```diff
@@ -290,14 +290,14 @@ def copy_run_image_metadata(self, old_run_label, new_run_label, **kwargs):
...
             new_image = response.json()[0]
-
             # Update the section and tissue of the copied image using
             # info from the original image
             updated_image = {
                 'section': item['section'] and item['section']['id'],
                 'tissue': item['tissue'] and item['tissue']['id'],
                 'project': item['section'] and \
                            item['section']['slide']['project']['id'],
+                'fov_size': item['fov_size'],
```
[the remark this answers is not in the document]
rfinck: Ah, the aperture change is in the upcoming release. We can do only the fov_size changes here, or we can make the aperture change in way that works both with the current and next version.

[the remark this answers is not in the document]
stanislaw-ionpath: I would go for the second option - "make the aperture change in way that works both with the current and next version". How do I do it? use `.get('aperture')`?

mibitracker/request_helpers.py
rfinck: Sorry, I was mistaken, either `fov_size` or `magnification` works as the parameter in the request, so your change is right,

mibitracker/request_helpers.py
```diff
@@ -290,15 +290,26 @@ def copy_run_image_metadata(self, old_run_label, new_run_label, **kwargs):
...
+            # TODO (Stanley) all these checks are for version compatibility.
+            # Should be removed for mibitracker >= V1.1.5
+            if 'aperture' in item:
+                updated_image.update({
+                    'aperture': item['aperture'] and item['aperture']['id']
+                    })
+            if 'imaging_preset' in item:
+                updated_image.update({'imaging_preset': item['imaging_preset']})
+            if 'lens1_voltage' in item:
+                updated_image.update({'imaging_preset': item['imaging_preset']})
```
rfinck: copy-paste error

mibitracker/request_helpers.py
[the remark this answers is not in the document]
rfinck: This looks good to me, once it's been testing on the demo I'll approve.

stanislaw-ionpath:
Rachel, I tested it on demo. It works.

The only exception is run 20180122_1242_1245 Point8 for which sed image is `sed.bmp` in the image metadata, but sed.bmp is missing.
```
 404 Client Error: Not Found for url: https://backend-dot-mibitracker-demo.appspot.com/download/?path=20180122_1242_1245-10008c4d-0bb3-42e2-8197-bd3969d67d2b%2FPoint8%2FRowNumber0%2FDepth_Profile0%2Fsed.bmp
```

PR #59, opened 2019-10-10
DPPA-479 Update mibitracker-client to specify project id when creating imageset

mapazarr:


mapazarr:
The tests were crashing because of `TypeError: attrib() got an unexpected keyword argument 'convert'`. I added `attrs` to the environment script in order to fix this issue.

mapazarr:
@rfinck I tested the code with the demo mibitracker and it works. This PR is ready for a review.

mibitracker/request_helpers.py
```diff
@@ -557,14 +557,16 @@ def get_channel_data(self, image_id, channel_name):
...
+    def create_imageset(self, image_ids, imageset_name, project_id,
                         imageset_description=None):
         """Creates a new imageset with the specified images.
 
         Args:
             image_ids: A list of ints of ids of the images in MibiTracker
                 corresponding to the images to be added to the new imageset.
             imageset_name: A string name for the new imageset.
+            project_id: An integer number for identifying the project to which
+                the images belong.
```
rfinck: The project id specifies where the _imageset_ belongs. The images must be either in that project, or in other projects to which the user has access and have sharing enabled. So the docstring should be something like "An integer id specifying the project in which to create the imageset. Note that all images in the set must be either in this project or shared with the user."

mapazarr: Done.

mapazarr:
@rfinck I updated the docstring with the information you requested. Please have a look and let me know if you would like me to change anything else.

mibitracker/request_helpers.py
```diff
@@ -557,14 +557,19 @@ def get_channel_data(self, image_id, channel_name):
...
         return skio.imread(buf)
 
-    def create_imageset(self, image_ids, imageset_name,
+    def create_imageset(self, image_ids, imageset_name, project_id,
                         imageset_description=None):
-        """Creates a new imageset with the specified images.
+        """Creates a new imageset in a project with the specified images.
+
+         Note that all images in the set must be either in the project specified
+         or in other projects to which the user has acces and sharing enabled.
```
rfinck: spelling: access

rfinck: also, maybe "sharing _is_ enabled" ?

mapazarr: Done.

PR #61, opened 2019-10-24
DPPA-460 Add timeout to request_helpers functions

stanislaw-ionpath:


mibitracker/request_helpers.py
```diff
@@ -562,7 +563,8 @@ def get_channel_data(self, image_id, channel_name):
             raise MibiTrackerError(
                 'Specified channel name is not present in image')
         buf = io.BytesIO()
-        response = requests.get(image_info['overlays'][channel_name])
+        response = requests.get(image_info['overlays'][channel_name],
+                                timeout=30)
```
rfinck: Let's have module-level constants for the mibitracker timeouts and the external url timeouts.

mibitracker/request_helpers.py
```diff
@@ -612,21 +614,31 @@ def _check_status(response):
         return response
 
     def get(self, *args, **kwargs):
+        if 'timeout' not in kwargs:
+            kwargs.update({'timeout': 10})
```
rfinck: I know it's only two lines but a simple helper function for this would be nice. Good call checking first so that the user can specify their own timeout if they want.

mibitracker/request_helpers.py
stanislaw-ionpath: done

mibitracker/request_helpers.py
stanislaw-ionpath: done

PR #63, opened 2019-10-30
DPPA-516 Update MibiImage metadata to MIBItiff 1.0 spec

mapazarr:
- change point_name [mibi.description on tiff] to fov_name
- add fov_id, which can be constructed from folder in old data
- allow extra kwargs in MibiImage constructor for user-defined attributes, and save them in the TIFF as well.
- maintain backwards compatibility with existing TIFFs


mapazarr:
This PR is ready for a review.

mibidata/mibi_image.py
```diff
@@ -110,7 +114,8 @@ def __init__(self, data, channels, run=None, date=None,
                  instrument=None, tissue=None, panel=None, version=None,
                  mass_offset=None, mass_gain=None, time_resolution=None,
                  miscalibrated=None, check_reg=None,
-                 datetime_format=_DATETIME_FORMAT, filename=None):
+                 datetime_format=_DATETIME_FORMAT, filename=None,
+                 optional_metadata=None):
```
jaytarolli: As we discussed offline, I think it would be better to continue adding attributes to the MibiImage class, such as description in this case. Then, instead of adding each as an optional argument to the constructor, implement kwargs and from the keywords match the values to attributes in the class.

jaytarolli:
FYI there is a description field saved as mibi.description in MIBITIFFs which maps to a MibiImage's point_name attribute. Should this PR address that and give MibiImage it's own description and point_name attributes? @rfinck @chris-brent

mibidata/mibi_image.py
```diff
@@ -56,14 +56,17 @@ class MibiImage():
             peak locations after mass recalibration.
         check_reg: Whether or not the maximum shift between depths is higher
             than a threshold.
+        filename: The name of the Run XML file which corresponds to the run
```
chris-brent: We're using JSON files for MIBIscope 1 :) It's also a convention that the filename corresponds to the run name, but it doesn't need to be that way.

mibidata/mibi_image.py
mapazarr: OK, I implemented `kwargs` for the metadata parameters in the constructor.

mapazarr:
The tests for the tiff module are currently failing in this PR because I'm updating the MibiImage class. I will fix them once I'm finished updating the class.

mibidata/tiff.py
```diff
@@ -311,6 +312,7 @@ def _page_metadata(page, description):
         'miscalibrated': description.get('mibi.miscalibrated'),
         'check_reg': description.get('mibi.check_reg'),
         'filename': description.get('mibi.filename'),
+        'optional_metadata': description.get('mibi.optional_metadata')
```
chris-brent: `'point_name': description.get('mibi.description')
`
on line 301 above can now be:
`'point_name': description.get('mibi.point_name')
`
?

mibidata/mibi_image.py
mapazarr: Fixed. It now reads `filename: The name of the file with the run metadata.`

mibidata/tiff.py
mapazarr: Fixed. Also, I'm using the recently introduced terminology: `fov_id` and `fov_name`, instead of `point_name`.

mapazarr:
@jaytarolli @rfinck @chris-brent This PR is ready for a review.
I added the new metadata fields and reorganized the old mixed up ones. I fixed the tiff module. I also added tests for backwards compatibility with the old metadata format.

mibidata/mibi_image.py
```diff
@@ -95,53 +114,79 @@ class MibiImage():
...
+        self.mass_gain = kwargs.pop('mass_gain')
+        self.time_resolution = kwargs.pop('time_resolution')
+        self.miscalibrated = kwargs.pop('miscalibrated')
+        self.check_reg = kwargs.pop('check_reg')
+        self.filename = kwargs.pop('filename')
+        self.description = kwargs.pop('description')
+
+        # whatever remains (if anything) is optional metadata
+        try:
+            optional_metadata = kwargs.pop('optional_metadata')
```
jaytarolli: I don't like the implementation of optional_metadata here because it requires two different ways of accessing MibiImage metadata, e.g.: `im.run` versus `im.optional_metadata['extra_field_name']`. In addition, when saved to a TIFF, the dictionary will just be saved as a single metadata entry where I think it would be nice if each key value pair in `optional_metadata` was saved as `mibi.key_name: value`.

image

I wonder if a solution to this could be saving a list of attribute names at the class level where it defaults to a copy of `_ATTRIBUTES` and adding to it any additional kwargs keys (so that the TIFF writer knows all of the metadata fields present). Then, `__setattr__` and `__getattr__`  could be overridden to allow a user to add his/her own metadata which would be saved and accessed in the same way as defined metadata fields.

mapazarr:
@rfinck @chris-brent @jaytarolli 
Shall we get rid of the `labels` property? The docstring says that it's deprecated.
https://github.com/ionpath/mibitracker-client/blob/d8366517ec79b2e242b4d26394fc2ad5ebc4ed95/mibidata/mibi_image.py#L168

mapazarr:
The force-push was only to amend the last commit's message, in order to specifically reflect that we increased the version to 1.0.

jaytarolli:
@mapazarr I can't think of any place we still use labels, but let's confirm that with @rfinck and @chris-brent to be safe.

rfinck:
> I can't think of any place we still use labels, but let's confirm that with @rfinck and @chris-brent to be safe.

Good call, we can get rid of it as part of moving to v1.0

mapazarr:
I added a TIFF test file to the repository saved with old metadata (prior to mibitiff v 1.0), in order to test that the software can read old files and keep backwards compatibility.

mapazarr:
@rfinck @jaytarolli @chris-brent Looking at the test for `mibi_image` and `tiff` I noticed the wide variety of ways to define the channels in `MibiImage`:
```python
STRING_LABELS = ('1', '2', '3')
TUPLE_LABELS = (('Mass1', 'Target1'), ('Mass2', 'Target2'), ('Mass3', 'Target3'))
MASS_LABELS = ('Mass1', 'Mass2', 'Mass3')
MASS_INTEGERS = (1, 2, 3)
TARGET_LABELS = ('Target1', 'Target2', 'Target3')
```
and the only way of defining them, if the image should be saved to a `TIFF`:
```python
CHANNELS = ((1, 'Target1'), (2, 'Target2'), (3, 'Target3'),
            (4, 'Target4'), (5, 'Target5'))
```
I also noticed that the format in `test_tiff` doesn't match any of the ones in `test_mibi_image`, though it is still accepted by a `MibiImage`.

Can we:
 1. get rid of the `MASS_LABELS` in `MibiImage`? (I don't  think they're used in the code.)
 2. use the same format for `TUPLE_LABELS` in `test_mibi_image` as the `CHANNELS` in `test_tiff`?

Code links for reference:
https://github.com/ionpath/mibitracker-client/blob/d8366517ec79b2e242b4d26394fc2ad5ebc4ed95/mibidata/tests/test_mibi_image.py#L21
https://github.com/ionpath/mibitracker-client/blob/d8366517ec79b2e242b4d26394fc2ad5ebc4ed95/mibidata/tests/test_tiff.py#L25

jaytarolli:
@mapazarr I think most of that is for compatibility with how channels/masses were and can be stored. There may be justification to simplify it, but it probably warrants a larger discussion and the changes specified by this ticket need to be in so that we can begin finalizing a new MIBITIFF specification.

mapazarr:
I implemented a new way to add the user-defined metadata key-value pairs as direct attributes to the class and a way to recover them to pass them further. I still need to adapt the tiff creation tools to the new format, so the tests are currently broken.

mibidata/mibi_image.py
mapazarr: @jaytarolli I tried adding a list of attribute names as an attribute to the class and overwriting the `setattr()` function in order to store user-defined attribute names in the list, but it's not working. The only way I managed to make this work is reaching to `self.__dict__` in order to get the names of the user-defined attributes. This way, it's possible to add them to the metadata and pass them further down in the code to i.e. the tiff creation. Please have a look and let me know if this works for you.

mapazarr:
@jaytarolli @chris-brent this is ready for a review. Once the format is accepted, I'd like to add a new test file to the repo with the new format and implement a unit test to read it and make sure everything works. Please let me know your thoughts to my comments above, especially the one regarding the software version from the tiff module.

mapazarr:
@jaytarolli @chris-brent @rfinck I should mention that the current implementation would capture anything that the user sets and is not a defined attribute of the image and put it into the metadata.
So if the user for instance produces a typo such as:
`image.runo = '20101010_1234'`
 there will be a `runo` attribute in the class and it will be saved as a metadata field, without any error.

rfinck:
I was assuming that we were allowing user-defined key-value pairs defined in the constructor to be saved in the TIFF, but not that arbitrary attributes added to an existing object would get saved. Also, the TIFF writer should fail if any of the required attributes are None.

mibidata/mibi_image.py
```diff
@@ -9,61 +9,84 @@
...
+                                 'check_reg', 'filename', 'description')
+# The attributes not to include in the metadata dictionary.
+_NON_METADATA_ATTRIBUTES = ('data', '_length', '_channels', 'masses', 'targets')
 
 
 class MibiImage():
     """A multiplexed image with labeled channels and metadata.
 
+    The format for setting the point name in the metadata has changed.
+    The correct format for setting the point name in the metadata is show in
```
jaytarolli: shown

mibidata/mibi_image.py
```diff
@@ -91,57 +115,66 @@ class MibiImage():
...
-        self.time_resolution = time_resolution
-        self.miscalibrated = miscalibrated
-        self.check_reg = check_reg
-        self.filename = filename
+        self.run = kwargs.pop('run', None)
+        self.coordinates = kwargs.pop('coordinates', None)
+        self.size = kwargs.pop('size', None)
+        self.slide = kwargs.pop('slide', None)
+        self.fov_id = kwargs.pop('fov_id', None)
+        self.fov_name = kwargs.pop('fov_name', None)
```
jaytarolli: One of the subtasks of the ticket was: "MibiImge constructor for backwards compatibility to change point_name keyword to fov_name" which I don't see implemented here.

mibidata/tests/test_mibi_image.py
```diff
@@ -40,6 +55,8 @@ def setUp(self):
         warnings.filterwarnings(
             'ignore',
             message='Anti-aliasing will be enabled by default.*')
+        self.maxDiff = None
```
jaytarolli: What is `maxDiff` for? I don't see it used anywhere.

mibidata/tests/test_mibi_image.py
```diff
@@ -95,14 +112,36 @@ def test_set_channels_invalid_tuple(self):
...
-        image.labels = TARGET_LABELS
-        self.assertEqual(image.channels, TARGET_LABELS)
+    @unittest.mock.patch('sys.stdout', new_callable=io.StringIO)
+    def test_convert_from_previous_metadata_versions(self, mock_stdout):
+        image = mi.MibiImage(TEST_DATA, TUPLE_LABELS)
+        image.point_name = OLD_METADATA['point_name']
+        image.folder = OLD_METADATA['folder']
+        image.version = OLD_METADATA['version']
+        image._convert_from_previous_metadata_versions()
+        message = ("WARNING! you are trying to use the old metadata "
```
jaytarolli: Single quotes in this file

mibidata/mibi_image.py
```diff
@@ -199,6 +214,37 @@ def _set_channels(self, channels, length):
                 'Channels must be a list of tuples of (int, str) or a '
                 'list of str')
 
+    def _convert_from_previous_metadata_versions(self):
+        """Convert old metadata format to new one.
+
+        This function ensures backwards compatibility for previous versions.
+        """
+        if self.version != MIBITIFF_VERSION:
+            print("WARNING! you are trying to use the old metadata "
```
jaytarolli: Single quotes in this file

mibidata/tests/test_tiff.py
```diff
@@ -10,7 +10,7 @@
 import warnings
 
 import numpy as np
-from skimage import io, img_as_ubyte, transform
+from skimage import io as skio, img_as_ubyte, transform
```
jaytarolli: Why was it necessary to change this?

mibidata/tests/test_tiff.py
```diff
@@ -25,15 +25,30 @@
...
     'folder': 'Point1/RowNumber0/Depth_Profile0',
     'aperture': '300um', 'instrument': 'MIBIscope1', 'tissue': 'Tonsil',
-    'panel': '20170916_1x', 'mass_offset': 0.1, 'mass_gain': 0.2,
-    'time_resolution': 0.5, 'miscalibrated': False, 'check_reg': False,
-    'filename': 'Run'
+    'panel': '20170916_1x', 'version': None, 'mass_offset': 0.1,
+    'mass_gain': 0.2, 'time_resolution': 0.5, 'miscalibrated': False,
+    'check_reg': False, 'filename': '20180703_1234_test'
 }
+OLD_TIFF_FILE = os.path.join(os.path.dirname(__file__),
```
jaytarolli: Why include a binary file for testing? Can't this be done in the same manner as the other tests where data and metadata defined here is saved to a temp file? We then can't track any changes between binary file versions in git and it makes it clear to the next person reading this test file what data/metadata exactly is being tested.

rfinck: Because if it's written using the old writer, we can't write it each time with the new one.

mapazarr: No we can't. Because we would be saving the file with the NEW metadata format, so we won't be able to test if we can read an old file with OLD metadata format, so any changes to the code would potentially break backwards compatibility and pass unnoticed.

mibidata/mibi_image.py
```diff
@@ -9,61 +9,84 @@
...
+            folder: The folder name for this image as determined by the
+                acquisition software.
+            dwell: A float pixel dwell time in :math:`ms`.
+            scans: A comma-separated list of image scan numbers.
+            aperture: A string name of the aperture used during image
+                acquisition.
+            instrument: A string identifier for the instrument used.
+            tissue: A string name of the tissue type.
+            panel: A string name of the panel used to stain the tissue.
+            version: A string identifier for the MIBItiff software version used.
```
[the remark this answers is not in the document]
rfinck: MIBItiff, to be consistent with our MIBIscope, MIBItracker, MIBIcontrol branding rather than TIFF format conventions.

mibidata/mibi_image.py
```diff
@@ -199,6 +214,37 @@ def _set_channels(self, channels, length):
...
+        """Convert old metadata format to new one.
+
+        This function ensures backwards compatibility for previous versions.
+        """
+        if self.version != MIBITIFF_VERSION:
+            print("WARNING! you are trying to use the old metadata "
+                  "format for setting the point name. If you are "
+                  "creating a new image, we recommend storing the "
+                  "point name as in the following example: "
+                  "'fov_id': 'Point1', 'fov_name': 'R1C3_Tonsil'.")
```
rfinck: If you are going to emit a warning, use `warnings.warn` not print.

Also, this one doesn't seem quite right -- it checks the version string only, states a recommendation, and then goes ahead and attempt to use that recommendation without confirming that there is a `'point_name'` attribute. Instead, you should probably check for `'point_name'` and if it's present and the version is old, state in the warning that it is being converted.

mibidata/mibi_image.py
```diff
@@ -91,57 +115,66 @@ class MibiImage():
         instrument: A string identifier for the instrument used.
         tissue: A string name of the tissue type.
         panel: A string name of the panel used to stain the tissue.
-        version: A string identifier for the software version used.
+        version: A string identifier for the MIBItiff software version used.
```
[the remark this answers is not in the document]
rfinck: Let's confirm that we are not using this for writing instrument version. If we are, or perhaps even if we aren't, we may want to specific`'MIBItiff_version'` so we can also specific instrument version etc.

mibidata/mibi_image.py
```diff
@@ -199,6 +214,37 @@ def _set_channels(self, channels, length):
...
+            try:
+                assert self.fov_id == self.folder.split('/')[0]
+            except AssertionError:
+                message = (f'The fov_id {self.fov_id} does not '
+                           f'match the folder {self.folder}.')
+                raise ValueError(message)
+        else:
+            message = (f'The fov_id {self.fov_id} does not '
+                       f'match the folder {self.folder}.')
+            raise ValueError(message)
```
rfinck: I think this can be simplified to:
```
if self.fov_id is None and self.folder is None:
  return
try:
  assert self.fov_id == self.folder.split('/')[0]
except (AttributeError, AssertionError):
  raise ValueError(...)

File: mibidata/mibi_image.py [hunk 16]
```diff
@@ -232,7 +278,16 @@ def __str__(self):
 
     def metadata(self):
         """Returns a dictionary of the image's metadata."""
-        return {key: getattr(self, key) for key in _ATTRIBUTES}
+        metadata_keys = [key for key in _EXPECTED_METADATA_ATTRIBUTES]
```
rfinck: `metadata_keys = list(_REQUIRED_METADATA_ATTRIBUTES)`

mibidata/mibi_image.py
```diff
@@ -232,7 +278,16 @@ def __str__(self):
...
+        # find user-defined metadata
+        metadata_keys.extend(self._user_defined_attributes())
+        return {key: getattr(self, key) for key in metadata_keys}
+
+    def _user_defined_attributes(self):
+        """Returns a list of keys corresponding to the user-defined metadata."""
+        return [key for key in self.__dict__
+                if key not in _EXPECTED_METADATA_ATTRIBUTES and
+                key not in _NON_METADATA_ATTRIBUTES]
```
rfinck: Instead of this, which as you've pointed out will include any attributes added to the instance after construction, I think we should store a dict of the kwargs remaining after popping the required ones and use that as the user-defined attributes. 

I'm also open to having a method where a user can add more attributes later, but I'd rather it be an explicit setter than just grabbing anything from the object. This could be something we do in a subsequent ticket.

mibidata/tiff.py
```diff
@@ -14,7 +14,7 @@
 from mibidata import mibi_image as mi, util
 
 # Increment this when making functional changes.
-SOFTWARE_VERSION = 'IonpathMIBIv0.2'
+SOFTWARE_VERSION = 'IonpathMIBIv1.0'
```
rfinck: Should we get this from the MibiImage rather than hard-coding it in two places?

mibidata/tiff.py
```diff
@@ -289,29 +275,39 @@ def _page_metadata(page, description):
...
+    """
+    try:
+        description['mibi.version']
+    except KeyError:
+        # if the key doesn't exist it means that the version was None and hence
+        # not saved in the tiff file
+        description['mibi.version'] = None
+
+    if description['mibi.version'] != mi.MIBITIFF_VERSION:
+        description['mibi.fov_name'] = description['mibi.description']
```
rfinck: Again here I think it's better to check on whether ``fov_name'` is present rather than relying on the version.

mibidata/tiff.py
```diff
@@ -93,27 +93,13 @@ def write(filename, image, sed=None, optical=None, ranges=None,
...
-        'mibi.time_resolution': getattr(image, 'time_resolution'),
-        'mibi.miscalibrated': getattr(image, 'miscalibrated'),
-        'mibi.check_reg': getattr(image, 'check_reg'),
-        'mibi.filename': getattr(image, 'filename'),
-    }
+    metadata = {'mibi.' + key: value for key, value in \
+                image.metadata().items() \
+                if key not in ['date']} # datetime objects are not
+                                        # JSON serializable
+
```
rfinck: The original idea here was the namespace the required attributes with the `'mibi.'` prefix, but now we are adding that to both them and user-defined extras. I'm not with the name-spacing at all at this point, but given that we've been using it maybe it's better we stick to that plan and don't add it to arbitrary attributes? Unfortunately we are not taking advantage of this in terms of how we convert TIFFs to/from MibiImages at this point but we can address that later.

mibidata/tiff.py
mapazarr: @rfinck I found that in the TIFF creator file, and wasn't sure about the purpose of it. I would also think that if we make a MibiImage version, the version of the TIFF should match it.  So, yes, I will definitely make a check on the MibiImage version and build the the software version accordingly. Or should it be exactly the same?
Alternatively, we can create a dedicated python file called mibitiff_version.py with the version number and methods to extract and compare version numbers and make all pieces of the software that need the version read it from there.

rfinck: The TIFFs we make need to be compatible with various third-party image analysis software software, and some of them use this tag to determine which reader to use. This includes but is not limited to BioFormats (general info [here](https://docs.openmicroscopy.org/bio-formats/5.8.2/formats/ionpath-mibi.html), code that uses version string [here](https://github.com/ome/bioformats/blob/develop/components/formats-gpl/src/loci/formats/in/IonpathMIBITiffReader.java#L54). So the MIBItiff version is a specific thing that has use and meaning outside of the library version. For now, let's not expand scope of this ticket and get the 1.0 version number in here in the simplest way that doesn't require it to be maintained separately from the number you've written into MibiImage (e.g. constructing this string by reading the number from there) and we can do other version management updates in a follow-up PR.

mibidata/mibi_image.py
mapazarr: I check for the old metadata format and convert to the new format. It's done in `_convert_from_previous_metadata_versions`.

mibidata/tests/test_mibi_image.py
mapazarr: It's for py.test to print full output. It's useful for debugging and it has no effect on the user. It was already defined in the tests for `tiff` so I activated it here as well.

mibidata/mibi_image.py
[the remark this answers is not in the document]
mapazarr: Ok, so I'm adding a new attribute `MIBItiff_version` alongside `version`.

mibidata/mibi_image.py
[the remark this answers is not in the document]
mapazarr: No change.

mibidata/tests/test_mibi_image.py
mapazarr: I'm using single quotes within the double quotes, so I'm keeping the double quotes. Let me know if you want another format for escaping the single quotes.

mibidata/mibi_image.py
mapazarr: Same as before.

mibidata/tests/test_tiff.py
mapazarr: I was trying to use the python `io` module, so I changed this one to `skio`. I didn't change it back because I thought it's clearer that it is not the standard python IO module. Let me know if you want it changed.

mibidata/mibi_image.py
```diff
@@ -9,61 +9,84 @@
...
+
+# The format of the run xml.
 _DATETIME_FORMAT = '%Y-%m-%dT%H:%M:%S'
 # The attributes to include in the metadata dictionary.
-_ATTRIBUTES = ('run', 'date', 'coordinates', 'size', 'slide', 'point_name',
-               'folder', 'dwell', 'scans', 'aperture',
-               'instrument', 'tissue', 'panel', 'version', 'mass_offset',
-               'mass_gain', 'time_resolution', 'miscalibrated', 'check_reg',
-               'filename')
+_EXPECTED_METADATA_ATTRIBUTES = ('run', 'date', 'coordinates', 'size', 'slide',
```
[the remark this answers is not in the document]
mapazarr: Done.

mibidata/mibi_image.py
```diff
@@ -9,61 +9,84 @@
...
-            string. Defaults to ``'%Y-%m-%dT%H:%M:%S'``.
-        mass_offset: Mass offset parameter used for mass calibration.
-        mass_gain: Mass gain used for mass calibration.
-        time_resolution: Parameter used for mass calibration.
-        miscalibrated: Whether or not there was significant difference between
-            peak locations after mass recalibration.
-        check_reg: Whether or not the maximum shift between depths is higher
-            than a threshold.
+        kwargs: A sequence of arguments that will be used to define the
+            metadata. A list of expected metadata keys follows; however, the
```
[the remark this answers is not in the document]
mapazarr: Done.

mibidata/mibi_image.py
```diff
@@ -81,7 +104,8 @@ class MibiImage():
             the image was acquired.
         size: A float size of the image width/height in :math:`\\mu m`.
         slide: A string or integer slide ID.
-        point_name: A string name for the point as assigned during the run.
+        fov_id: A string identifying the point in the run, such as 'Point2'.
```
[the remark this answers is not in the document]
mapazarr: Ok, updated.

mibidata/mibi_image.py
mapazarr: Ok, I restructured the function and updated the warning. In general, I think we should enforce the version check, rather than relying on a check of possible attributes, especially since we are allowing the user to define his own attributes.

mibidata/mibi_image.py
mapazarr: Done, the way I had it before was because there is a case in which fov_id can be None but folder not, which is for the old metadata format (using point_name), but I reorganized the check of the old metadata to happen before the check on the new convention, so now it works with the simplified logic.

mibidata/mibi_image.py
mapazarr: Done.

mibidata/tiff.py
mapazarr: Ok, I made the software version dependent on the MIBItiff version.

mibidata/tiff.py
mapazarr: @rfinck I think in this case we should rely on the version, and assume the user is familiar with the required keywords of the current version. Otherwise, after a few versions, it will be very hard to follow the logic of this function, with multiple if and try checks on different variables.
I implemented unit tests to ensure backwards compatibility. What we would need at some point is a better way to compare MIBItiff versions, but this can be done in a separate PR. Let me please know if you still want me to change this.

mibidata/tiff.py
mapazarr: Ok, but I need to tag the user-defined metadata somehow, otherwise it's very hard to locate them in the tiff, when reading a tiff file and retrieving the metadata attributes for the MibiImage. I'm using now `user-def.` to tag the user-defined attributes, in order not to mix the attributes with the required ones, marked as `mibi.`.

rfinck: Can't you grab everything else from the dict that doesn't have the prefix / isn't listed as required? Similar to how you already construct a MibiImage from kwargs?

mibidata/mibi_image.py
mapazarr: @rfinck I understand your concern; but I implemented the dictionary on the 1st place, and @jaytarolli didn't like the solution. I now implemented a solution that I think should solve all conflicts:
 - all remaining kwargs (after popping the required attributes) are now added via `setattr` to the `MibiImage` instance, so that they can be retrieved as usual `image.personalized_attr`.
 - at the same time, I created a protected member `_user_defined_attributes` to store a list of keys in order to keep track of the user-defined attributes, otherwise it's very hard to pass them further to the tiff creator.
Let me know if this doesn't work for you.

mapazarr: EDIT: please consider that pylint now complains if you try to access the user-defined metadata, for example:
`Instance of 'MibiImage' has no 'mass_range' member (no-member)`.

mibidata/mibi_image.py
```diff
@@ -91,57 +115,66 @@ class MibiImage():
...
+        self.instrument = kwargs.pop('instrument', None)
+        self.tissue = kwargs.pop('tissue', None)
+        self.panel = kwargs.pop('panel', None)
+        self.mass_offset = kwargs.pop('mass_offset', None)
+        self.mass_gain = kwargs.pop('mass_gain', None)
+        self.time_resolution = kwargs.pop('time_resolution', None)
+        self.miscalibrated = kwargs.pop('miscalibrated', None)
+        self.check_reg = kwargs.pop('check_reg', None)
+        self.filename = kwargs.pop('filename', None)
+        self.description = kwargs.pop('description', None)
```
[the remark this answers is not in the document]
mapazarr: We could do something like 
```
        for attr in _REQUIRED_METADATA_ATTRIBUTES:
            if attr not in ['date']:
                setattr(self, attr, kwargs.pop(attr, None))
```
but then, we get a few `pylint` errors of the type `(no-member)`. I think that pylint enforces a list of required attributes to be specified, so I would leave it as-is, unless you would like me to change it.

mibidata/mibi_image.py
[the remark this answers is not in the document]
mapazarr: Done.

mibidata/mibi_image.py
```diff
@@ -9,61 +9,84 @@
...
+            run: A string name of the run during which this image was acquired.
+            date: The run date. It can either be a datetime object, or a string.
+                If a string, it will be parsed according to the
+                `datetime_format``.
+            coordinates: A tuple of (x, y) stage coordinates at which the image
+                was acquired; stage coordinates should be in microns.
+            size: A float size of the image width/height in  :math:`\\mu m`.
+            slide: A string or integer slide ID.
+            fov_id: A string identifying the point in the run, such as 'Point2'.
+            fov_name: A string name for the FoV as assigned during the run.
```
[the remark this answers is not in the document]
mapazarr: Done.

mibidata/mibi_image.py
```diff
@@ -81,7 +104,8 @@ class MibiImage():
             the image was acquired.
         size: A float size of the image width/height in :math:`\\mu m`.
         slide: A string or integer slide ID.
-        point_name: A string name for the point as assigned during the run.
+        fov_id: A string identifying the point in the run, such as 'Point2'.
+        fov_name: A string name for the FoV as assigned during the run.
```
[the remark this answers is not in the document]
mapazarr: Done.

mibidata/mibi_image.py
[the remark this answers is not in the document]
mapazarr: @rfinck @jaytarolli @chris-brent Can we rename the old `version` to something like `instrumet_version`? Or have it inside a dictionary `instrument`, with keys `name` and `version`? Otherwise it can be confusing/misleading for the user to see 2 attributes called version.

mapazarr:
Ok, this PR is ready for another review. I have some concerns though:

- MibiImage attribute `version` is misleading in several ways:
  - as a new user I would think it's the version of the image (i.e. v 0.1 without bg correction, v 1.0 with bg correction)
  - if I read the string doc I would think it's the software version used: `version: A string identifier for the software version used.`
  - if I see it next to MIBItiff_version I get confused about why do we need 2 'sofware' versions for 1 file.
  - if I see it's value (i.e. 'alpha') it seems to refer to the instrument version rather than the software version.
So I would strongly recommend to rename it and update the stringdoc.
- fail tiff writing if required metadata is None is not backwards-compatible friendly: if I read an old file and try to write a new tiff out of it, i cannot since the requirements used for writing the file in previous versions are less strict, and fields such as `description` and `version`, among others, can be empty. We can still ask for stronger requirements now, but we are forcing the user to get those attributes filled out of eventually insuficient information.

Please let me know your thoughts here.

mibidata/tiff.py
mapazarr: I will have a look, but I think there might be more info stored there.

mapazarr: OK I implemented the latter: I add a list of user-defined attribute keywords to the TIFF metadata, that makes the retrieval of the user-defined attributes straightforward, without the need for a special namespace prefix, similar to what I implemented for MibiImage.

rfinck:
@jaytarolli @mapazarr I pushed some changes that I hope can be the final ones. My goal was to keep this to its initial scope, which was:

- change point_name [mibi.description on tiff] to fov_name
- add fov_id, which can be constructed from folder in old data
- allow extra kwargs in MibiImage constructor for user-defined attributes, and save them in the TIFF as well.
- maintain backwards compatibility with existing TIFFs

We can save changes to how we report instrument, software, etc. versions as a MibiImage attribute for another day.

Please let me know if you see any issues, I will read it through one more time tomorrow as well.

mibidata/tests/test_tiff.py
```diff
@@ -248,6 +249,8 @@ def test_sort_channels_before_writing(self):
 
         tiff.write(self.filename, unordered_image)
         image = tiff.read(self.filename)
+        print(image.metadata())
+        print(self.image.metadata())
```
mapazarr: Here again, can we remove the printouts?

mibidata/tiff.py
```diff
@@ -285,47 +288,47 @@ def _page_metadata(page, description):
...
 
 
-def _convert_from_previous_metadata_versions(description):
-    """Convert old metadata format to new one.
+def _convert_from_previous(description):
+    """Convert old metadata format for backwards compatibility.
 
-    This function ensures backwards compatibility for previous versions.
+    Most of these conversions would happen during MibiImage construction,
+    but we do them here in case reading the info only.
```
mapazarr: This docstring is misleading. This function is necessary for reading an old TIFF file (written with previous versions of the TIFF creator and the MibiImage class) and converting its metadata to the new format.

mibidata/tiff.py
```diff
@@ -285,47 +288,47 @@ def _page_metadata(page, description):
...
+        warnings.warn(
+            'The "fov_id" attribute is now required if "folder" is '
+            'specified. Setting "fov_id" to {}.'.format(
+                description['mibi.fov_id']))
+    if (not description.get('mibi.folder') and description.get('mibi.fov_id')
+            and description.get('mibi.fov_id').startswith('FOV')):
+        description['mibi.folder'] = description['mibi.fov_id']
+        warnings.warn(
+            'The "folder" attribute is required if "fov_id" is specified. '
+            'Setting "folder" to {}.'.format(description['mibi.folder']))
```
mapazarr: This logic is a bit complicated to follow. The previous implementation based on the MIBITIFF version was cleaner.

mibidata/tests/test_tiff.py
```diff
@@ -220,21 +219,23 @@ def test_open_file_with_old_metadata(self):
...
                                                '%Y-%m-%dT%H:%M:%S'),
-            'description': None})
+        })
+        del expected['description']
         del expected['version']
+        for key, val in metadata.items():
+            print(key)
+            if val != expected.get(key):
+                print(val, expected.get(key))
         self.assertEqual(metadata, expected)
```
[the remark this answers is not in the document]
rfinck: Yes, thanks, that was leftover from debugging.

mibidata/tests/test_tiff.py
rfinck: Will do.

mibidata/tiff.py
rfinck: If reading a TIFF into a MibiImage, everything after that TODO is unnecessary and can be deleted, because the old metadata gets passed to the MibiImage constructor which does the conversion. My first draft relied on that, but then I realized (or rather a test caught) that then the `tiff.info` function returns unconverted data so this was my workaround to that.

mibidata/tiff.py
rfinck: I don't love this section of the code because it feels redundant to what happens during construction of a MibiImage, but relying on the user to say `MIBItiff_version=1.0` is less robust than looking at the actual parameters passed and converting them to the current version if necessary.

mibidata/tests/test_mibi_image.py
```diff
@@ -112,16 +109,11 @@ def test_set_channels_invalid_tuple(self):
         with self.assertRaises(ValueError):
             image.channels = invalid_tuple_3
 
-    def test_convert_from_previous_metadata_versions(self):
-        image = mi.MibiImage(TEST_DATA, TUPLE_LABELS)
-        image.point_name = OLD_METADATA['point_name']
-        image.folder = OLD_METADATA['folder']
-        image.MIBItiff_version = None
+    def test_convert_from_previous(self):
```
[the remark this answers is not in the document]
rfinck: Good point that it's redundant, but the previous version added attributes to the MibiImage instance and then had to workaround the fact that they were inconsistent with what was established when constructing the instance. Since `_convert_from_previous` is a private method I'd rather test it in the way it actually gets used (as part of the constructor) rather than in an artificial way. We can remove this test if it's a duplicate at this point.

rfinck:
@mapazarr I disagree on the first point and agree on the second, which I indicated in my previous comment should be done as a follow-up since it extends beyond this class to how we improve version coordination across this repository and others .

It is important to note that every time we create a MibiImage with this version of the code, we will be creating one that is of _this_ version. I do not want to have to manually pass in a version number to the constructor to say "hey, these are old parameters, please convert for me", or "these are the 1.0 parameters, I promise" but maybe I made a mistake and they're not. The parameters should speak for themselves. 

Reading in a _file_ is a particular case of this, since a file exists and persists outside of this class or library. The MIBItiff files must say what version they are, and their version number can be of one those parameters that speak for themselves mentioned above.

mibidata/mibi_image.py
```diff
@@ -118,16 +110,14 @@ class MibiImage():
...
-            version used.
         mass_offset: Mass offset parameter used for mass calibration.
         mass_gain: Mass gain used for mass calibration.
         time_resolution: Parameter used for mass calibration.
         miscalibrated: Whether or not there was significant difference
             between peak locations after mass recalibration.
         check_reg: Whether or not the maximum shift between depths is higher
             than a threshold.
-        filename: The name of the file with the run metadata.
+        filename: The name of the file from the iwith the run metadata.
```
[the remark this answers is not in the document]
rfinck: I think your comment was on an intermediate commit; I cleaned this up already.

mibidata/tiff.py
```diff
@@ -289,29 +286,49 @@ def _page_metadata(page, description):
...
+    Most of these conversions would happen during MibiImage construction,
+    but we do them here in case reading the info only.
+    """
+    if not description.get('mibi.fov_name') and description.get(
+            'mibi.description'):
+        description['mibi.fov_name'] = description.pop('mibi.description')
+    # TODO: Clean up repetition between this and the same MibiImage method
+    if description.get('mibi.folder') and not description.get('mibi.fov_id'):
+        description['mibi.fov_id'] = description['mibi.folder'].split('/')[0]
+        warnings.warn(
```
jaytarolli: I think I've seen this with other libraries when printing warnings, so I don't know if it's necessarily a bug here:

image

jaytarolli:
Just to chime in on the versioning here: My feeling is that the versioning of the MibiImage class is handled already with release tags. If a user is going to write code that relies or assumes a set of particular attributes in that class, then it is on he/she to either use a compatible release of this repository or heed the warnings printed and update his/her code.

mapazarr:
@rfinck In the way I implemented the versioning of the MibiImage, the user does not have to provide the version number; if it's left unset, the check version function was performing a check for the old metadata and if not found, it was assuming that the version is the current one, defined in the MIBITIFF global variable. I know there is a bit of ambiguity in that check, because there was no software tag before. But I still think it would be more maintainable if we start using versions and enforcing them from 1.0 onwards. We can assume that the user is aware of the specifics for the current version, by providing sufficient documentation.

And if at some point, if for instance we're at version 2.0 and he wants to use format 1.0 he can still do so, but specify the version in the initialisation, instead of having the convert function guess the version that the user is attempting to use.

But ok, I see you are not happy with this solution, so let's move on.

mibidata/tiff.py
mapazarr: This is a warning to the user to inform that he's trying to read old data, but we're converting it to the current format.

mibidata/tests/test_mibi_image.py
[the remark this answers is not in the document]
mapazarr: I know, I was mocking the construction in order to specifically test the function. Usually each function (even if it's protected) should have its test, otherwise, if something changes in the constructor (i.e. a loop around the convert function), it could be easily missed by the test and pass unnoticed, diminishing the coverage and leaving an important part of the software untested.

Issue #64, opened 2019-11-08
Can mibitracker-client visualization works with the Json profile from MIBI tracker?


YubinXie:
Hi I am trying to tune the visualization profile from MIBI tracker and then visualize the image in local with the same color profile. Is there a quick way to do this? Thanks!

rfinck:
Yes! Use `mibidata.color` -- there's an example about midway down the notebook here: https://github.com/ionpath/mibitracker-client/blob/master/MibiImage_Tutorial.ipynb
and more documentation here:
https://ionpath.github.io/mibitracker-client/mibidata.html#mibidata.color.composite

YubinXie:
I check the content in the json file and the composite function. They do not seem match. For example, in the Json file, it has blur, brightness information:
        "channels": {
            "CD8": {
                "blur": 0,
                "brightness": 0.55,
                "channel": "CD8",
                "color": "Yellow",
                "hex": "#FFFF00",
                "intensity_higher": 15,
                "intensity_lower": 1
            },

However, for the composite function, it only takes gamma, and min_scale...

rfinck:
Yes, I realized that right after my first comment and edited it above to say that we'll add an example of mapping from the json shortly. It's a good question and we should have it in there.

YubinXie:
Sounds great. Thank you!

YubinXie:
@rfinck 
one more thing, gray color is missing from color composite when I compare the color choice in MIBI tacker and MIBI client. I'd like to use gray as an option in MIBI client color composite...

rfinck:
Got it -- thanks for the heads up, it needs to be handled a little differently but we can add it in.

rfinck:
@YubinXie Code is in, we haven't issued a new release yet so you might need to git clone/pull to get the function that you can see here: https://github.com/ionpath/mibitracker-client/blob/master/mibidata/color.py#L230

YubinXie:
Sorry for the late reply. This function works on the MIBI tracker color profile! Thanks!

Though the composite function still has no 'Gray' option. Will it be possible to add it? Thanks!

rfinck:
Hi Yubin,
Glad to hear you're using it!
The previous change addressed making the json file work including any 'Gray' channels there, but I just added 'Gray' to the accepted list in #86. If you've got additional requests, feel free to open up a new issue!

PR #65, opened 2019-11-12
DPPA-556: Method to use mibitracker vis setting json to make composite in mibitracker-client

stanislaw-ionpath:
compose_overlay_from_image_data similar to the one used in mibitracker.

### DIFF
--- a/mibidata/color.py
+++ b/mibidata/color.py
@@ -227,7 +227,7 @@ def composite(image, color_map, gamma=1/3, min_scaling=10):
     return np.uint8(screened * 255)
 
 
-def compose_overlay_from_image_data(image, overlay_settings):
+def compose_overlay(image, overlay_settings):
     """Overlays multiple image channels using overlay_settings from mibitracker.
 
     Args:
--- a/mibidata/color.py
+++ b/mibidata/color.py
@@ -242,6 +242,19 @@ def gray2hsl(array, angle):
 
 def compose_overlay_from_image_data(image, overlay_settings):
     """Overlays multiple image channels using overlay_settings from mibitracker.
+
+    Args:
+        image: A MibiImage.
+        overlay_settings: Color overlay dictionary with the following fields:
+            'brightness': float between -1 and 1.
+            'channel': Name of the MibiImage channel.
+            'color': One of the following: 'Cyan', 'Yellow', 'Magenta', 'Green',
+                'Orange', 'Violet', 'Red', 'Blue', or 'Gray'.
+            'intensity_higher': Upper limit of the channel intensity.
+            'intensity_lower': Lower limit of the channel intensity.
+
+    Returns:
+        An NxMx3 uint8 array of an RGB image.
     """
     # Convolve with this filter no matter what, to mimic browser rendering.
     kernel = np.array([
@@ -269,7 +282,7 @@ def compose_overlay_from_image_data(image, overlay_settings):
             hsl = gray2hsl(float_array, COLORS[item['color']])
             rgb = hsl2rgb(hsl)
         if i == 0:
-            compose_overlay = rgb
+            overlay = rgb
         else:
-            compose_overlay = _porter_duff_screen(composite, rgb)
-    return compose_overlay
+            overlay = _porter_duff_screen(composite, rgb)
+    return np.uint8(overlay * 255)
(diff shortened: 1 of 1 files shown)


mibidata/color.py
```diff
@@ -224,3 +225,51 @@ def composite(image, color_map, gamma=1/3, min_scaling=10):
...
+    """Converts NxN grayscale to RGB of a single color.
+
+    The input array is assumed to be scaled in the unit interval [0, 1]
+    The angle is in the range [0, 2*pi]
+    """
+    hsl = np.zeros((array.shape[0], array.shape[1], 3))
+    hsl[:, :, 0] = angle
+    hsl[array > 0, 1] = 1
+    hsl[:, :, 2] = array / 2
+    return hsl
```
rfinck: This is already a helper function in this file.

stanislaw-ionpath: Oh,  yes there is!

rfinck:
Fixes #64

mibidata/color.py
```diff
@@ -224,3 +225,51 @@ def composite(image, color_map, gamma=1/3, min_scaling=10):
...
+    Returns:
+        An NxMx3 uint8 array of an RGB image.
+    """
+    # Convolve with this filter no matter what, to mimic browser rendering.
+    kernel = np.array([
+        [0.05, 0.1, 0.05],
+        [0.1, 0.4, 0.1],
+        [0.05, 0.1, 0.05]
+    ])
+    for i, item in enumerate(overlay_settings):
```
rfinck: This does not work directly with the json as downloaded from mibitracker, which has the form:

```
{
  'image_id': {
    'channels': {
      'channel': {'brightness': float, 'color': color, ...}
    }
  }
}
```

You can either make it work with this for any image id with something like 

```
[settings] = overlay_settings.values()
for i, (channel, item) in enumerate(settings['channels'].items()):
  int_array = image[channel]
  ...
```

or expect the user to have selected the the `'channels'` already, in which case you still need to iterate not only through the key.

mibidata/color.py
```diff
@@ -224,3 +225,51 @@ def composite(image, color_map, gamma=1/3, min_scaling=10):
...
+        np.clip(float_array, 0, 1, out=float_array)
+        if item['color'] == 'Gray':
+            rgb = np.stack((float_array, float_array, float_array), axis=2)
+        else:
+            hsl = _gray2hsl(float_array, COLORS[item['color']])
+            rgb = hsl2rgb(hsl)
+        if i == 0:
+            overlay = rgb
+        else:
+            overlay = _porter_duff_screen(composite, rgb)
```
rfinck: `composite` is the name of another function in this file, I think you meant `overlay`

stanislaw-ionpath:
@rfinck I made changes and tested it on mibitracker images. I think it is ok now. Please have a look.

mibidata/color.py
```diff
@@ -224,3 +225,114 @@ def composite(image, color_map, gamma=1/3, min_scaling=10):
...
+            Each channel should have the following fields:
+                'color': One of the following: 'Cyan', 'Yellow', 'Magenta',
+                    'Green', 'Orange', 'Violet', 'Red', 'Blue', or 'Gray'.
+                'brightness': (optional) float between -1 and 1; defaults to 0.
+                'intensity_higher': (optional) Upper limit of the channel
+                    intensity; defaults to maximum counts in the channel.
+                'intensity_lower': (optional) Lower limit of the channel
+                    intensity; defaults to 0,
+                'blur': (optional) float between 0 and 1 defining the gaussian
+                    blur of the channel; defaults to 0.
```
rfinck: The formatting of this doesn't work in the sphinx docs, and I think it can be simplified to reference what one gets from mibitracker rather than reproducing it in its entirety. If you do want to use what you write here in the error message below, save it as a string rather than copy-pasting.

Here's what it looks like now:
image

stanislaw-ionpath:
@rfinck now it looks like this:
image

mibidata/color.py
```diff
@@ -224,3 +225,83 @@ def composite(image, color_map, gamma=1/3, min_scaling=10):
             gamma)
     screened = _screen(data_map)
     return np.uint8(screened * 255)
+
+
+def compose_overlay(image, overlay_settings):
+    """Overlays multiple image channels using overlay_settings from mibitracker.
```
rfinck: MIBItracker

mibidata/color.py
```diff
@@ -224,3 +225,83 @@ def composite(image, color_map, gamma=1/3, min_scaling=10):
...
+
+    The overlay_settings are intended to have the form of a channels.json file
+    as downloaded from mibitracker but they can have any of the of the following
```
rfinck: MIBItracker

stanislaw-ionpath:
@rfinck done

PR #66, opened 2019-11-18
Fix conjugates order for panels with multiple targets on the same mass

stanislaw-ionpath:


mibidata/panels.py
```diff
@@ -8,6 +8,10 @@
 from mibidata import util
 
 
+def pascal_case(st):
```
murataksoy: Doc?

mibidata/panels.py
```diff
@@ -8,6 +8,10 @@
...
+    return ''.join(x for x in st.title() if x.isalnum())
+
```
murataksoy: Is this OK?:
```
In [5]: aaa="β-tubulin"                                                                                                                   

In [6]: aaa                                                                                                                               
Out[6]: 'β-tubulin'

In [7]: aaa.title()                                                                                                                       
Out[7]: 'Β-Tubulin'

```

murataksoy: Ah, I see now what you are doing - this is just about "target" and "mass". OK

mibidata/panels.py
rfinck: Should this be a private function?

PR #67, opened 2019-11-21
MIC-1203 Validate and convert metadata formats in setters.

rfinck:
This follow-up to DPPA-516 allows the validation and conversion of old metadata formats to work when set directly on a MibiImage instance, e.g. `image.point_name = 'R3C1'`. It also better separates the validation for each attribute.

### DIFF
--- a/mibidata/mibi_image.py
+++ b/mibidata/mibi_image.py
@@ -161,6 +161,7 @@ class MibiImage():
 
     @property
     def point_name(self):
+        """Return fov_name instead of deprecated point_name."""
         return self.fov_name
 
     @point_name.setter
--- a/mibidata/mibi_image.py
+++ b/mibidata/mibi_image.py
@@ -168,7 +168,7 @@ class MibiImage():
         """Convert deprecated point_name to fov_name."""
         warnings.warn('The "point_name" attribute is deprecated. '
                       'Setting "fov_name" to "{}".'.format(value))
-        self.fov_name = value
+        self.fov_name = value  # pylint: disable=attribute-defined-outside-init
 
     @property
     def folder(self):
@@ -184,8 +184,6 @@ class MibiImage():
                     'The "fov_id" attribute is now required if "folder" is '
                     'specified. Setting "fov_id" to {}.'.format(fov))
                 self._fov_id = fov
-                self._folder = value
-                return
             elif self.fov_id != fov:
                 raise ValueError('fov_id must match folder, but here '
                                  'folder={} and you are trying to set fov_id '
@@ -203,14 +201,12 @@ class MibiImage():
             warnings.warn(
                 'The "folder" attribute is required if "fov_id" is specified. '
                 'Setting "folder" to {}.'.format(value))
-            self._fov_id = value
             self._folder = value
         elif self.folder and value != self.folder.split('/')[0]:
             raise ValueError('fov_id must match folder, but here '
                              'folder={} and you are trying to set fov_id '
                              'to {}.'.format(self.folder, value))
-        else:
-            self._fov_id = value
+        self._fov_id = value
 
     @property
     def channels(self):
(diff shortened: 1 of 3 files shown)


mibidata/mibi_image.py
```diff
@@ -157,11 +159,54 @@ def __init__(self, data, channels, **kwargs):
...
 
-        # check version for backwards compatibility
-        self._convert_from_previous()
+    @property
+    def point_name(self):
+        return self.fov_name
+
+    @point_name.setter
+    def point_name(self, value):
+        """Convert deprecated point_name to fov_name."""
```
rileysparsons: Great.

mibidata/mibi_image.py
```diff
@@ -157,11 +159,54 @@ def __init__(self, data, channels, **kwargs):
...
-        # check consistency between fov_id and folder
-        self._check_fov_id()
+    @property
+    def fov_id(self):
+        return self._fov_id
+
+    @fov_id.setter
+    def fov_id(self, value):
+        """Enforce consistency with folder."""
+        if value and not self.folder:
```
rileysparsons: 👍

mibidata/tests/test_mibi_image.py
```diff
@@ -203,10 +200,10 @@ def test_metadata_backwards_compatibility(self):
                                                       mi._DATETIME_FORMAT)
         metadata['fov_name'] = metadata['point_name']
         metadata['fov_id'] = metadata['folder'].split('/')[0]
-        del metadata['point_name']
         metadata['description'] = None
         self.assertEqual(image.metadata(), metadata)
-        self.assertEqual(len(image._user_defined_attributes), 0)
+        # point_name is retained as user-defined
+        self.assertEqual(image._user_defined_attributes, ['point_name'])
```
mapazarr: I think we should try to get rid of the `point_name`, otherwise it can be confusing to have twice the same information in the MibiImage instance.
I think we can easily get rid of `point_name` in the setter. I will test a possible solution for this.

rfinck: I think we should leave it in since we are allowing arbitrary user-defined attributes it would be unexpected to exclude this particular string as one of those. Also users may have code that relies on this from the version; this gives them a warning to update without breaking anything immediately.

mapazarr:
@rfinck @rileysparsons I implemented a fix to remove the deprecated point_name from user-defined attributes after converting to fov_name. I also updated the docstring.

mibidata/mibi_image.py
```diff
@@ -157,11 +159,54 @@ def __init__(self, data, channels, **kwargs):
             setattr(self, k, v)
             self._user_defined_attributes.append(k)
 
-        # check version for backwards compatibility
-        self._convert_from_previous()
+    @property
+    def point_name(self):
+        return self.fov_name
```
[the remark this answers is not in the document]
mapazarr: Done.

mibidata/tests/test_mibi_image.py
mapazarr: I just uploaded the fix for this, shall I undo it?
Sorry, I just read the message.

mibidata/mibi_image.py
```diff
@@ -169,6 +169,7 @@ def point_name(self, value):
         warnings.warn('The "point_name" attribute is deprecated. '
                       'Setting "fov_name" to "{}".'.format(value))
         self.fov_name = value  # pylint: disable=attribute-defined-outside-init
+        self._user_defined_attributes.remove('point_name')
```
rfinck: This will raise a ValueError is a user tries to set `point_name` on an existing MibiImage. I am going to remove it as part of a cleanup commit that also fixes some docstring formatting.

mapazarr: Oh, ok, yes I realized later I should have used a try-except block in the setter. Anyways, no need for it now.

mapazarr:
I know we merged this, but I was thinking that keeping the old `point_name` attribute as user-defined attribute will still break the code of users relying on this attribute, since we are not storing it into the `mibi.` namespace within the tiff metadata. We might want to rethink this issue.

PR #70, opened 2019-11-23
DPPA-565: Default to save MIBItiff as float

jaytarolli:
Part 1 of this ticket is to have the TIFF writer by default save the underlying dtype unless specified otherwise.

### DIFF
--- a/mibidata/tests/test_tiff.py
+++ b/mibidata/tests/test_tiff.py
@@ -16,7 +16,7 @@ from skimage.external.tifffile import TiffFile
 from mibidata import mibi_image as mi
 from mibidata import tiff, util
 
-DATA = np.arange(500).reshape(10, 10, 5).astype(np.float)
+DATA = np.arange(500).reshape(10, 10, 5).astype(np.float32)
 SED = np.arange(100).reshape(10, 10)
 CAROUSEL = np.random.randint(0, 255, (4096, 4096, 3), np.uint8)
 X_COORD, Y_COORD = tiff._BOTTOM_LABEL_COORDINATES
@@ -90,7 +90,7 @@ class TestWriteReadTiff(unittest.TestCase):
         self.assertIsNone(sed)
         self.assertIsNone(optical)
         self.assertIsNone(label)
-        self.assertEqual(image.data.dtype, np.float)
+        self.assertEqual(image.data.dtype, np.float32)
 
     def test_default_ranges(self):
         tiff.write(self.filename, self.float_image)
@@ -265,10 +265,10 @@ class TestWriteReadTiff(unittest.TestCase):
 
             np.testing.assert_equal(np.squeeze(tif.data), DATA[:, :, i])
             self.assertTupleEqual(tif.channels, (CHANNELS[i], ))
-            self.assertEqual(tif.data.dtype, np.float)
+            self.assertEqual(tif.data.dtype, np.float32)
 
     def test_tiff_dtype_correct_arguments(self):
-        supported_dtypes = ['float', 'uint16', np.float, np.uint16]
+        supported_dtypes = ['float', 'uint16', np.float32, np.uint16]
         for dtype in supported_dtypes:
             tiff.write(self.filename, self.float_image, multichannel=True,
                        dtype=dtype)
@@ -287,33 +287,33 @@ class TestWriteReadTiff(unittest.TestCase):
     def test_write_float_from_float_tiff_dtype_none(self):
         tiff.write(self.filename, self.float_image, multichannel=True)
         image = tiff.read(self.filename)
-        self.assertEqual(image.data.dtype, np.float)
+        self.assertEqual(image.data.dtype, np.float32)
         np.testing.assert_equal(
-            image.data, self.float_image.data.astype(np.float))
+            image.data, self.float_image.data.astype(np.float32))
 
     def test_write_float_from_float_dtype_float(self):
         tiff.write(self.filename, self.float_image, multichannel=True,
                    dtype='float')
         image = tiff.read(self.filename)
-        self.assertEqual(image.data.dtype, np.float)
+        self.assertEqual(image.data.dtype, np.float32)
         np.testing.assert_equal(
-            image.data, self.float_image.data.astype(np.float))
+            image.data, self.float_image.data.astype(np.float32))
 
     def test_write_float_from_float_dtype_np_float(self):
         tiff.write(self.filename, self.float_image, multichannel=True,
-                   dtype=np.float)
+                   dtype=np.float32)
         image = tiff.read(self.filename)
-        self.assertEqual(image.data.dtype, np.float)
+        self.assertEqual(image.data.dtype, np.float32)
         np.testing.assert_equal(
-            image.data, self.float_image.data.astype(np.float))
+            image.data, self.float_image.data.astype(np.float32))
 
     def test_write_float_from_int_dtype_float(self):
         tiff.write(self.filename, self.int_image, multichannel=True,
                    dtype='float')
         image = tiff.read(self.filename)
-        self.assertEqual(image.data.dtype, np.float)
+        self.assertEqual(image.data.dtype, np.float32)
         np.testing.assert_equal(
-            image.data, self.float_image.data.astype(np.float))
+            image.data, self.float_image.data.astype(np.float32))
 
     def test_write_int_from_int_dtype_none(self):
         tiff.write(self.filename, self.int_image, multichannel=True)
--- a/mibidata/tests/test_tiff.py
+++ b/mibidata/tests/test_tiff.py
@@ -268,7 +268,7 @@ class TestWriteReadTiff(unittest.TestCase):
             self.assertEqual(tif.data.dtype, np.float)
 
     def test_tiff_dtype_correct_arguments(self):
-        supported_dtypes = ['float', 'int', np.float, np.uint16]
+        supported_dtypes = ['float', 'uint16', np.float, np.uint16]
         for dtype in supported_dtypes:
             tiff.write(self.filename, self.float_image, multichannel=True,
                        dtype=dtype)
@@ -339,6 +339,15 @@ class TestWriteReadTiff(unittest.TestCase):
         self.assertEqual(image.data.dtype, np.uint16)
         np.testing.assert_equal(
             image.data, self.float_image.data.astype(np.uint16))
+
+        lossy_image = mi.MibiImage(DATA + 0.001, CHANNELS, **METADATA)
+        with self.assertWarns(UserWarning):
+            tiff.write(self.filename, lossy_image, multichannel=True,
+                       dtype='uint16')
+        image = tiff.read(self.filename)
+        np.testing.assert_equal(
+            image.data, self.float_image.data.astype(np.uint16))
+
 
 
 if __name__ == '__main__':
--- a/mibidata/tests/test_tiff.py
+++ b/mibidata/tests/test_tiff.py
@@ -348,12 +348,9 @@ class TestWriteReadTiff(unittest.TestCase):
 
     def test_write_float_as_uint16_warns(self):
         lossy_image = mi.MibiImage(DATA + 0.001, CHANNELS, **METADATA)
-        with self.assertWarns(UserWarning):
+        with self.assertRaises(ValueError):
             tiff.write(self.filename, lossy_image, multichannel=True,
                        dtype='uint16')
-        image = tiff.read(self.filename)
-        np.testing.assert_equal(
-            image.data, self.float_image.data.astype(np.uint16))
 
 
 if __name__ == '__main__':
--- a/mibidata/tests/test_tiff.py
+++ b/mibidata/tests/test_tiff.py
@@ -268,12 +268,12 @@ class TestWriteReadTiff(unittest.TestCase):
             self.assertEqual(tif.data.dtype, np.float32)
 
     def test_tiff_dtype_correct_arguments(self):
-        supported_dtypes = ['float', 'uint16', np.float32, np.uint16]
+        supported_dtypes = [np.float32, np.uint16]
         for dtype in supported_dtypes:
             tiff.write(self.filename, self.float_image, multichannel=True,
                        dtype=dtype)
 
-        unsupported_types = ['abcdef', np.str, np.bool]
+        unsupported_types = ['abcdef', np.str, np.bool, np.uint8, np.float64]
         for dtype in unsupported_types:
             with self.assertRaises(ValueError):
                 tiff.write(self.filename, self.float_image, multichannel=True,
@@ -284,22 +284,14 @@ class TestWriteReadTiff(unittest.TestCase):
             tiff.write(self.filename, self.int_image, write_float=True)
             tiff.write(self.filename, self.float_image, write_float=False)
 
-    def test_write_float_from_float_tiff_dtype_none(self):
+    def test_write_float32_from_float32_tiff_dtype_none(self):
         tiff.write(self.filename, self.float_image, multichannel=True)
         image = tiff.read(self.filename)
         self.assertEqual(image.data.dtype, np.float32)
         np.testing.assert_equal(
             image.data, self.float_image.data.astype(np.float32))
 
-    def test_write_float_from_float_dtype_float(self):
-        tiff.write(self.filename, self.float_image, multichannel=True,
-                   dtype='float')
-        image = tiff.read(self.filename)
-        self.assertEqual(image.data.dtype, np.float32)
-        np.testing.assert_equal(
-            image.data, self.float_image.data.astype(np.float32))
-
-    def test_write_float_from_float_dtype_np_float(self):
+    def test_write_float32_from_float32_dtype_np_float32(self):
         tiff.write(self.filename, self.float_image, multichannel=True,
                    dtype=np.float32)
         image = tiff.read(self.filename)
@@ -307,30 +299,22 @@ class TestWriteReadTiff(unittest.TestCase):
         np.testing.assert_equal(
             image.data, self.float_image.data.astype(np.float32))
 
-    def test_write_float_from_int_dtype_float(self):
+    def test_write_float32_from_int_dtype_np_float32(self):
         tiff.write(self.filename, self.int_image, multichannel=True,
-                   dtype='float')
+                   dtype=np.float32)
         image = tiff.read(self.filename)
         self.assertEqual(image.data.dtype, np.float32)
         np.testing.assert_equal(
             image.data, self.float_image.data.astype(np.float32))
 
-    def test_write_int_from_int_dtype_none(self):
+    def test_write_uint16_from_uint16_dtype_none(self):
         tiff.write(self.filename, self.int_image, multichannel=True)
         image = tiff.read(self.filename)
         self.assertEqual(image.data.dtype, np.uint16)
         np.testing.assert_equal(
             image.data, self.float_image.data.astype(np.uint16))
 
-    def test_write_int_from_int_dtype_uint16(self):
-        tiff.write(self.filename, self.int_image, multichannel=True,
-                   dtype='uint16')
-        image = tiff.read(self.filename)
-        self.assertEqual(image.data.dtype, np.uint16)
-        np.testing.assert_equal(
-            image.data, self.float_image.data.astype(np.uint16))
-
-    def test_write_int_from_int_dtype_np_uint16(self):
+    def test_write_uint16_from_uint16_dtype_np_uint16(self):
         tiff.write(self.filename, self.int_image, multichannel=True,
                    dtype=np.uint16)
         image = tiff.read(self.filename)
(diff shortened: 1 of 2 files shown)


rfinck:
I want to wait to merge this until we have a couple dependencies in elsewhere.

mibidata/tiff.py
```diff
@@ -82,12 +84,17 @@ def write(filename, image, sed=None, optical=None, ranges=None,
...
-        range_dtype = 'I'
+    if write_dtype and not write_dtype in ['float', 'int', np.float, np.uint16]:
+        raise ValueError('Invalid dtype specification.')
+    if not write_dtype:
+        range_dtype = 'I' if\
+            np.issubdtype(image.data.dtype, np.integer) else 'd'
     else:
-        range_dtype = 'd'
+        range_dtype = 'I' if write_dtype in ['int', np.uint16] else 'd'
+
```
mapazarr: If you end up doing this, you're not respecting the type-choices you promised before. Especially for int types, you pre-define 'int' and np.uint16 as accepted types, but you end up using int. Similar with float.

mibidata/tiff.py
```diff
@@ -132,10 +139,10 @@ def write(filename, image, sed=None, optical=None, ranges=None,
                 max_value = (341, range_dtype, 1, ranges[i][1])
                 page_tags = coordinates + [page_name, min_value, max_value]
 
-                if write_float:
-                    to_save = image.data[:, :, i].astype(np.float32)
+                if range_dtype == 'I':
+                    to_save = image.data[:, :, i].astype(np.uint16)
                 else:
-                    to_save = image.data[:, :, i]
+                    to_save = image.data[:, :, i].astype(np.float)
```
mapazarr: Here again, you're bypassing the choices you gave in the list and end up using only either `np.uint16` if it's integer type or `np.float` if it's float type.

mibidata/tiff.py
```diff
@@ -179,22 +186,17 @@ def write(filename, image, sed=None, optical=None, ranges=None,
...
             with TiffWriter(target_filename,
                             software=SOFTWARE_VERSION) as infile:
 
-                if write_float:
-                    to_save = image.data[:, :, i].astype(np.float32)
+                if range_dtype == 'I':
+                    to_save = image.data[:, :, i].astype(np.uint16)
                 else:
-                    to_save = image.data[:, :, i]
+                    to_save = image.data[:, :, i].astype(np.float)
```
mapazarr: Same thing here. Plus, we don't need to have 2 places to convert the data. Please do a check at the beginning of the file and convert it there.

jaytarolli:
@mapazarr A quick response to the overarching theme of your review: I believe we want to restrict the data type saved as a MIBItiff to either uint16 or float. I know that in the past we had issues trying to support uint8 and uint16 simultaneously and eventually decided to go with just uint16. The rationale behind a lot of this is that the MIBItiff needs to be supported across a number of IONpath and third party applications and, in my opinion, the best way to do so is to restrict the data to just these two types. So the goal of this PR is to generalize the TIFF writer a little bit so that calling code (such as MIBI/O) can define how to save the data.

mapazarr:
@jaytarolli I agree that for cross-platform support the best way would be to be very specific with the types used. Would it make sense then to store the data as either `np.uint64` and `np.float64`?

As for the use-cases where one could have signed integer values, I guess we could then store the tiff as a float. Can we implement a check a check in the code at the point, where we convert the data, that throws a warning in case one tries to convert signed integer values into unsigned ones and then defaults to float?

jaytarolli:
What would be the rationale for using uint64 and float64? Specifically for integer counts, 65k should be plenty of room to hold pixel values without worry of data loss. Changing to those data types would require changes in MIBItracker in addition to this and I'm not sure it adds anything we're currently lacking.

As for converting signed integers, I'm of the mindset that the TIFF writer should limit the data types saved as a MIBItiff file to those that are supported, and it's on the user to properly convert the data beforehand if needed. That way the user is making the decision on how to deal with a potential loss of data or precision.

mibidata/tiff.py
```diff
@@ -64,16 +64,18 @@ def write(filename, image, sed=None, optical=None, ranges=None,
...
-            opening properly in certain software such as Halo). Defaults to
-            False which will save the image data as uint16. Note: setting
-            write_float to True does not normalize or scale the data before
-            saving, however saves the integer counts as floating point numbers.
+        write_dtype: Forces the image data saved as either float or uint16. Can
+            specify `'float'` or `np.float` to force data to be saved as
+            floating point values or `'int'` or `np.uint16` to save as integer
+            values. Defaults to None, which saves data as its original type.
+            Note that forcing native float image data to uint16 could result
+            in a loss of precision as values are clipped.
```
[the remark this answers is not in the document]
rfinck: Our default TIFF generation will be float (in a pipeline outside of this library), and this PR enables that to persist as the default type since the TIFF writer will not alter that type unless specified by the user. 

That said, we are supporting float and uint16, not general int types.

mibidata/tiff.py
```diff
@@ -64,16 +64,18 @@ def write(filename, image, sed=None, optical=None, ranges=None,
...
 
     Raises:
         ValueError: Raised if the image is not a
             ``mibitof.mibi_image.MibiImage`` instance, or if its coordinates,
-            size, masses or targets are None.
+            size, masses or targets are None, or if `write_dtype` is not one
+            of 'float', 'int', np.float, or np.uint16.
```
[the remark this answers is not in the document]
rfinck: The formatting you suggest 
```
ValueError: Raised if
    - the image is not a :class:`mibitof.mibi_image.MibiImage` instance
    - the image coordinates, size, masses or targets are None.	
    - `write_dtype` is not an accepted type.
```
does not work well with our sphinx docs, I think it's fine as written now but if it needs to change please refer to how to correctly format bullet points or similar.

mibidata/tiff.py
rfinck: This is for a specific TIFF tag, not the data itself, and needs to conform to what's allowed for that tag.

mibidata/tiff.py
```diff
@@ -82,12 +84,17 @@ def write(filename, image, sed=None, optical=None, ranges=None,
         raise ValueError('Image coordinates and size must not be None.')
     if image.masses is None or image.targets is None:
         raise ValueError('Image channels must contain both masses and targets.')
-    if np.issubdtype(image.data.dtype, np.integer) and not write_float:
-        range_dtype = 'I'
+    if write_dtype and not write_dtype in ['float', 'int', np.float, np.uint16]:
```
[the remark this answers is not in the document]
rfinck: Because the only int type we support is uint16, we cannot only check if an int subtype. I think if the user requests uint16 we should be checking if it can be losslessly converted and warning if not (an easy way to do this is to convert and then check equality with `np.all(np.equal(...))`).

mibidata/tiff.py
```diff
@@ -46,7 +46,7 @@ def _cm_to_micron(arg):
 
 # pylint: disable=too-many-branches,too-many-statements
 def write(filename, image, sed=None, optical=None, ranges=None,
-          multichannel=True, write_float=False):
+          multichannel=True, write_dtype=None):
```
rfinck: Let's just call it `dtype`, not `write_dtype`? And maybe keep `write_float` and either give a deprecation warning or error in case a user tries it?

mibidata/tiff.py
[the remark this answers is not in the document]
jaytarolli: I just tried a number of different approaches and can confirm that I could not get a list to generate under the ValueError.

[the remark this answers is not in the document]
jaytarolli: Never mind, one more shot and I got it

mibidata/tests/test_tiff.py
```diff
@@ -262,15 +265,75 @@ def test_write_single_channel_tiffs(self):
...
             self.assertTupleEqual(tif.channels, (CHANNELS[i], ))
-            self.assertEqual(tif.data.dtype, np.uint16)
+            self.assertEqual(tif.data.dtype, np.float)
+
+    def test_tiff_dtype_correct_arguments(self):
+        supported_dtypes = ['float', 'int', np.float, np.uint16]
+        for dtype in supported_dtypes:
+            tiff.write(self.filename, self.float_image, multichannel=True,
+                       write_dtype=dtype)
+
```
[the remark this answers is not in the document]
jaytarolli: I believe that is covered in tests later which save based different combinations of dtype and actual data type, where the a round trip image is compared back to the data saved. This test just verifies that the argument restrictions are properly enforced.

mibidata/tiff.py
```diff
@@ -82,12 +84,17 @@ def write(filename, image, sed=None, optical=None, ranges=None,
...
+        range_dtype = 'I' if\
+            np.issubdtype(image.data.dtype, np.integer) else 'd'
     else:
-        range_dtype = 'd'
+        range_dtype = 'I' if write_dtype in ['int', np.uint16] else 'd'
+
     if ranges is None:
-        ranges = [(0, m) for m in image.data.max(axis=(0, 1))]
+        dtype_conversion = int if range_dtype == 'I' else float
+        ranges = [(0, dtype_conversion(m)) for m in image.data.max(axis=(0, 1))]
```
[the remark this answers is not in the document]
jaytarolli: Yes, otherwise the add tag function in the underlying TIFF writer complains the value is not of the correct type. Not sure how it worked before this change...

mibidata/tests/test_tiff.py
```diff
@@ -262,15 +265,75 @@ def test_write_single_channel_tiffs(self):
...
+        unsupported_types = ['abcdef', np.str, np.bool]
+        for dtype in unsupported_types:
+            with self.assertRaises(ValueError):
+                tiff.write(self.filename, self.float_image, multichannel=True,
+                           write_dtype=dtype)
 
     def test_write_float_tiff(self):
```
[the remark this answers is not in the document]
jaytarolli: I've split it into individual tests with descriptive names

mibidata/tests/test_tiff.py
```diff
@@ -262,15 +265,75 @@ def test_write_single_channel_tiffs(self):
...
+
+        tiff.write(self.filename, self.int_image, multichannel=True,
+                   write_dtype=np.uint16)
+        image = tiff.read(self.filename)
+        self.assertEqual(image.data.dtype, np.uint16)
+        np.testing.assert_equal(
+            image.data, self.float_image.data.astype(np.uint16))
+
+        tiff.write(self.filename, self.float_image, multichannel=True,
+                   write_dtype='int')
```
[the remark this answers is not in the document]
jaytarolli: I added a warning if the data to be saved after conversion is different than the input data, per @rfinck's suggestion.

mibidata/tiff.py
jaytarolli: Done. Raises an error since it would be hard to tell if `write_float` was actually specified or if it was set to the default `False` value.

mibidata/tiff.py
jaytarolli: Done.

mibidata/tiff.py
```diff
@@ -82,12 +89,26 @@ def write(filename, image, sed=None, optical=None, ranges=None,
...
+    if dtype and not dtype in ['float', 'uint16', np.float, np.uint16]:
+        raise ValueError('Invalid dtype specification.')
+    if not dtype:
+        range_dtype = 'I' if\
+            np.issubdtype(image.data.dtype, np.integer) else 'd'
     else:
-        range_dtype = 'd'
+        range_dtype = 'I' if dtype in ['uint16', np.uint16] else 'd'
+
+    save_dtype = np.uint16 if range_dtype == 'I' else np.float
```
rfinck: There was a compatibility reason we were using `np.float32` rather than `float` which is float64. Please retain the use of np.float32 here. Moreover, I think
- we should raise an exception not just a waning if the data isn't equal after converting
- this entire section can be simplified to something like 

```
    if dtype == 'float':
        save_dtype = np.float32
        range_dtype = 'd'
    elif dtype == 'uint16':
        save_dtype = np.uint16
        range_dtype = 'I'
    elif np.issubdtype(image.data.dtype, np.floating):
        save_dtype = np.float32
        range_dtype = 'd'
    else:
        save_dtype = np.uint16
        range_dtype = 'I'

    to_save = image.data.astype(save_dtype)
    if not np.all(np.equal(to_save, image.data)):
        raise ValueError('Cannot convert data from '
                         f'{image.data.dtype} to {save_dtype}')

    if ranges is None:
        ranges = [(0, m) for m in to_save.max(axis=(0, 1))]

File: mibidata/tests/test_tiff.py [hunk 13]
```diff
@@ -16,7 +16,7 @@
 from mibidata import mibi_image as mi
 from mibidata import tiff, util
 
-DATA = np.arange(500).reshape(10, 10, 5).astype(np.uint16)
+DATA = np.arange(500).reshape(10, 10, 5).astype(np.float)
```
rfinck: Change this to `np.float32` which will require some changes below too.

mibidata/tiff.py
```diff
@@ -64,16 +64,23 @@ def write(filename, image, sed=None, optical=None, ranges=None,
...
-            opening properly in certain software such as Halo). Defaults to
-            False which will save the image data as uint16. Note: setting
-            write_float to True does not normalize or scale the data before
-            saving, however saves the integer counts as floating point numbers.
+        dtype: Forces the image data saved as either float or uint16. Can
+            specify `'float'` or `np.float` to force data to be saved as
+            floating point values or `'uint16'` or `np.uint16` to save as
+            unsigned 16-bit integer values. Defaults to None, which saves data
+            as its original type. Note that forcing native float image data to
+            uint16 could result in a loss of precision as values are clipped.
```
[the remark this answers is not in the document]
jaytarolli: Actually, I think the easier/cleaner way for users to specify would be using the numpy type, so if anything I would remove the string names. I included them because I know plenty of times I am working with MIBItiffs but haven't actually imported numpy in that particular file. If anything, I would say simplify and leave just `np.uint16` and `np.float32`. Do you agree?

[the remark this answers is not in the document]
rfinck: I'm a little torn, because while in general I agree it's best to specific types by the types themselves, in this case we have two strict types we use when writing (np.uint16 and np.float32), but we are happy to convert from broader related types if not losing data (e.g. save as uint16 from np.uint8 or np.int if in [0, 2**16), save as np.float32 if float64). And, as you point out, using the specific types requires numpy to be imported. But since this parameter only gets used when a user wants to _force_ the dtype, perhaps it's best to be explicit as to the exact type used so they can think through the consequences. I'd be more hesitant if this were a required parameter, because then having to specific exact dtype would feel like an excessive burden to the user.

So, after all that, if your suggestion is to use `np.uint16` and `np.float32`, I'm fine with that. 

One last option is that we define constants based on these, so the available options to the user are from our module, and we also use these throughout our code for simplicity.

```
FLOAT = np.float32
UINT = np.uint16
```

then dtype is one of `tiff.FLOAT, tiff.UINT`, ...

mibidata/tiff.py
jaytarolli: I like that this forces the data into one of the two types and leaves it on the user to decide how to convert the data before writing the TIFF so that there is no room for confusion later. I elected to keep the check in that `dtype` is one of the allowed to avoid the case where a user specifies `np.uint8` or `abcdef` and not realize why the data is trying to (and failing to be) converted into uint16.

mibidata/tiff.py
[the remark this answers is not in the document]
jaytarolli: I think I would prefer using `np.uint16` and `np.float32` so that, like you say, it's explicit to the user which dtype is being saved. I'll restrict the allowed arguments to just those.

mibidata/tiff.py
rfinck: 100% agreed that there should be a check first that they specify an allowed dtype, my comment didn't make this clear but was intended to be after that check.

mibidata/tiff.py
```diff
@@ -64,16 +64,23 @@ def write(filename, image, sed=None, optical=None, ranges=None,
...
+        write_float: Deprecated, will raise ArgumentError if specified. To
```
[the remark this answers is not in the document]
jaytarolli: Oops.

mibidata/tiff.py
```diff
@@ -64,16 +64,23 @@ def write(filename, image, sed=None, optical=None, ranges=None,
...
+            argument instead.
 
     Raises:
-        ValueError: Raised if the image is not a
-            ``mibitof.mibi_image.MibiImage`` instance, or if its coordinates,
-            size, masses or targets are None.
+        ValueError: Raised if
+
+            * The image is not a ``mibitof.mibi_image.MibiImage`` instance
+            * Ccoordinates, size, masses or targets are None
```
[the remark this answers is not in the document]
jaytarolli: Done (also linked to the class).

mibidata/tiff.py
```diff
@@ -64,16 +64,23 @@ def write(filename, image, sed=None, optical=None, ranges=None,
...
+            specify the dtype of the saved image, please use the `dtype`
+            argument instead.
 
     Raises:
-        ValueError: Raised if the image is not a
-            ``mibitof.mibi_image.MibiImage`` instance, or if its coordinates,
-            size, masses or targets are None.
+        ValueError: Raised if
+
+            * The image is not a ``mibitof.mibi_image.MibiImage`` instance
```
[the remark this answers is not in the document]
jaytarolli: I used :class: to link to the class documentation and not have it formatted as the constructor. Also updated mibitof to mibidata throughout the file.

mibidata/tiff.py
```diff
@@ -64,16 +64,23 @@ def write(filename, image, sed=None, optical=None, ranges=None,
...
+            * Ccoordinates, size, masses or targets are None
+            * `dtype` is not one of 'float', 'uint16', np.float, or np.uint16
```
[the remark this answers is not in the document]
jaytarolli: Done.

mibidata/tests/test_tiff.py
jaytarolli: Done.

jaytarolli:
@rfinck Bouncing back to you.

mibidata/tests/test_tiff.py
```diff
@@ -262,15 +265,93 @@ def test_write_single_channel_tiffs(self):
...
+        unsupported_types = ['abcdef', np.str, np.bool, np.uint8, np.float64]
+        for dtype in unsupported_types:
+            with self.assertRaises(ValueError):
+                tiff.write(self.filename, self.float_image, multichannel=True,
+                           dtype=dtype)
+
+    def test_write_float_is_deprecated(self):
+        with self.assertRaises(ValueError):
+            tiff.write(self.filename, self.int_image, write_float=True)
+            tiff.write(self.filename, self.float_image, write_float=False)
```
rfinck: This only tests that one of these statements raises. And I think testing once is fine, but am noting that having the second statement doesn't do anything.

mibidata/tiff.py
```diff
@@ -64,30 +64,60 @@ def write(filename, image, sed=None, optical=None, ranges=None,
...
-        write_float: If True, saves the image data as float32 values (for
-            opening properly in certain software such as Halo). Defaults to
-            False which will save the image data as uint16. Note: setting
-            write_float to True does not normalize or scale the data before
-            saving, however saves the integer counts as floating point numbers.
+        dtype: Forces the image data saved as either float or uint16. Can
+            specify or ``np.float32`` to force data to be saved as floating
+            point values or ``np.uint16`` to save as unsigned 16-bit integer
+            values. Defaults to None, which saves data as its original type, if
+            the original type can converted without a loss of data.
```
rfinck: This wording still needs to be improved. I'd suggest

dtype: One of `(np.float32, np.uint16)` to force the the dtype of the saved image data.  Defaults to `None`, which chooses the format based on the data's input type, and will convert to `np.float32` or `np.uint16` from other float or int types, respectively, if it can do so without a loss of data.

mibidata/tiff.py
```diff
@@ -64,30 +64,60 @@ def write(filename, image, sed=None, optical=None, ranges=None,
...
-            size, masses or targets are None.
+        ValueError: Raised if
+
+            * The image is not a :class:`mibidata.mibi_image.MibiImage` instance
+            * The :class:`mibidata.mibi_image.MibiImage` coordinates, size,
+              masses or targets are None
+            * `dtype` is not one of ``np.float32`` or ``np.uint16``
+            * `write_float` has been specified.
+            * Converting the native :class:`mibidata.mibi_image.MibiImage` dtype
+              to the specified or inferred ``dtype`` results in a loss of data.
```
rfinck: Please be consistent with whether you end your items with `.` or not

mibidata/tiff.py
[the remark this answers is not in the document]
rfinck: I still don't think you need to do the conversion here if you use the data you've already converted:

```
if ranges is None:
  ranges = [(0, m) for m in to_save.max(axis=(0, 1))]

mibidata/tests/test_tiff.py
jaytarolli: Good catch. The second one actually fails when tested individually because I only tested `if write_float` and not that it isn't None. Since I know I have places in code where I pass `write_float=False`, I am making sure that each case is tested.

mibidata/tiff.py
jaytarolli: I am embarrassed for myself...

mibidata/tiff.py
jaytarolli: Done.

mapazarr:
Ok, now it makes more sense, having only 2 specific supported types, one for float and one for int. Thanks!

PR #72, opened 2019-12-05
[MIC-1221] Fix fov_id and folder setters when None; add aperture constraint

rileysparsons:


mibidata/tests/test_mibi_image.py
```diff
@@ -127,6 +127,20 @@ def test_check_fov_id(self):
         with self.assertRaises(ValueError):
             image.fov_id = None
 
+    def test_check_fov_id_without_folder(self):
+        metadata = METADATA.copy()
+        metadata.pop('folder')
+        image = mi.MibiImage(TEST_DATA, TUPLE_LABELS, **metadata)
+        self.assertEqual(image.fov_id, 'Point1');
```
chris-brent: I'm not the only one who adds ; in python!!!!!! My People!!!!

mibidata/tests/test_mibi_image.py
```diff
@@ -127,6 +127,20 @@ def test_check_fov_id(self):
...
+        self.assertEqual(image.folder, 'Point1');
```
chris-brent: ; ;)

mibidata/tests/test_mibi_image.py
```diff
@@ -127,6 +127,20 @@ def test_check_fov_id(self):
...
+
+    def test_check_folder_without_fov_id(self):
+        metadata = METADATA.copy()
+        metadata.pop('fov_id')
+        image = mi.MibiImage(TEST_DATA, TUPLE_LABELS, **metadata)
+        self.assertEqual(image.fov_id, 'Point1');
```
chris-brent: ;

mibidata/tests/test_mibi_image.py
```diff
@@ -127,6 +127,20 @@ def test_check_fov_id(self):
...
+        self.assertEqual(image.folder, 'Point1/RowNumber0/Depth_Profile0');
```
chris-brent: ;

mibidata/mibi_image.py
```diff
@@ -154,7 +154,14 @@ def __init__(self, data, channels, **kwargs):
             self.date = date
 
         for attr in SPECIFIED_METADATA_ATTRIBUTES[1:]:
-            setattr(self, attr, kwargs.pop(attr, None))
+            if attr == 'folder':
```
chris-brent: 👍

rileysparsons:
@rfinck:

My tests are both passing for the following reason:

I only attempt to set `folder` and `fov_id` if they exist in `kwargs`.

```
            if attr == 'folder':
                if 'folder' in kwargs:
                    self.folder = kwargs.pop(attr)
            elif attr == 'fov_id':
                if 'fov_id' in kwargs:
                    self.fov_id = kwargs.pop(attr)
```

The tests I've written then fail when I remove this logic because we set all of the `SPECIFIED_METADATA_ATTRIBUTES` regardless of whether they are in `kwargs`. If they do not exist they are set to `None`.

mibidata/tests/test_mibi_image.py
```diff
@@ -127,6 +127,20 @@ def test_check_fov_id(self):
         with self.assertRaises(ValueError):
             image.fov_id = None
 
+    def test_check_fov_id_without_folder(self):
+        metadata = METADATA.copy()
+        metadata.pop('folder')
+        image = mi.MibiImage(TEST_DATA, TUPLE_LABELS, **metadata)
+        self.assertEqual(image.fov_id, 'Point1');
+        self.assertEqual(image.folder, None); # We allow folder to not be provided in the constructor kwargs, it is set to None in this case.
```
rileysparsons: @rfinck - if you alter constructor behavior (i.e. fix that bug) this line will likely fail. Just pointing that out.

rfinck: Thanks @rileysparsons -- I reverted these tests to the original ones you added which correctly caught this issue, and updated the setters to prevent it. Note that now if you try `image.folder = None` it will leave the folder as-is rather than break consistency with fov_id, and I think that is better behavior.

mibidata/mibi_image.py
```diff
@@ -19,7 +19,12 @@
                                  'time_resolution', 'miscalibrated',
                                  'check_reg', 'filename', 'description',
                                  'version')
-
+APERTURE_MAP = {
```
chris-brent: I'm fine to have this defined in one location, and this might well be the best place for it as we already use mibitracker-client in the mibitracker backend. But we should use one definition **everywhere** and mibi_shared defines it as:
APERTURE_1MM = u'A'
APERTURE_300UM = u'B'
APERTURE_100UM = u'C'
APERTURE_30UM = u'D'
APERTURES = (
    (APERTURE_1MM, u'1 mm'),
    (APERTURE_300UM, u'300 \u03BCm'),
    (APERTURE_100UM, u'100 \u03BCm'),
    (APERTURE_30UM, u'30 \u03BCm'),
)
Note the unicode and '1 mm' not '1mm'

rileysparsons:
> Can you do this with a setter rather than checking every item in this loop?

Good call.

mibidata/mibi_image.py
rileysparsons: Done.

rileysparsons:
@rfinck - I did end up having to hack something together for the `test_open_file_with_old_metadata` to pass. But again, since we are just testing whether we can correctly parse a TIFF file's known metadata, I think it should be acceptable.

mibidata/mibi_image.py
```diff
@@ -200,14 +213,25 @@ def fov_id(self):
...
         self._fov_id = value
 
+    @property
+    def aperture(self):
+        return self._aperture
+
+    @aperture.setter
+    def aperture(self, value):
+        if value is not None and value not in APERTURE_MAP.values():
+            raise ValueError('Invalid aperture code \'{}\', must use {} for {}, respectively'.format(value, ', '.join(list(APERTURE_MAP.values())), ', '.join(list(APERTURE_MAP.keys()))))
```
rfinck: I think it's easier to read if you print the actual dictionary, something like `'Invalid aperture \'{}\', must use values from the following map: {}'.format(APERTURE_MAP).`

rileysparsons: OK.

mibidata/mibi_image.py
```diff
@@ -200,14 +213,28 @@ def fov_id(self):
...
+    @property
+    def aperture(self):
+        return self._aperture
+
+    @aperture.setter
+    def aperture(self, value):
+        if value is not None and value not in APERTURE_MAP.values():
+            raise ValueError(
+                'Invalid aperture code \'{}\', must use values \
+                    from the following map: {}'.format(value, APERTURE_MAP)
```
rfinck: Nit: You don't need the `\' or extra indentation for the line break inside the parentheses:
```
raise ValueError(
    'Invalide code .... '
    'from the following ...'
)

rileysparsons: 🤯
Did not know that. Thanks.

rileysparsons:
@rfinck - merge at your convenience. Happy to tag if I have permissions.

PR #73, opened 2019-12-06
Modfiy the circleci build process so that the build will fail if the git tag and setup.py do not match

chityala-ion:


.circleci/config.yml
```diff
@@ -87,13 +87,37 @@ jobs:
...
+    docker:
+      - image: circleci/python:3.7 # container for the build job
 
+    steps:
+      - checkout:
+          path: ~/mibitracker-client
+      - run:
+          name: Check if tag and setup.py are consistent
+          command: |
+            python check_tag.py
```
murataksoy: I assume you confirmed that this works without any python environment or additional libraries, correct?

check_tag.py
```diff
@@ -0,0 +1,59 @@
...
+                    stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+    stdout, _ = proc.communicate()
+    version = stdout.strip().decode('utf-8')
+    pattern = r"v(\d+).(\d+).(\d+)(.*?)"
+    match = re.match(pattern, version)
+    if match:
+        major = match.group(1)
+        minor = match.group(2)
+        build = match.group(3)
+        return "v"+".".join((major, minor, build))
```
murataksoy: For this and the next function, why do you need to split into minor-major-build? Don't you want to check check if the tag names are the same, regardless of their format ?

check_tag.py
```diff
@@ -0,0 +1,59 @@
...
+                    major = match.group(1)
+                    minor = match.group(2)
+                    build = match.group(3)
+                    return "v"+".".join((major, minor, build))
+    return ""
+
+if __name__ == '__main__':
+    if not check_tag():
+        raise \
+            Exception("The version number in setup.py and git tag do not match")
```
murataksoy: Can you also write some extra comments to guide the user in this case, e.g., "please change bla bla line in setup.py to -print current version here- " ?

.circleci/config.yml
chityala-ion: Yes. It works on github.

check_tag.py
chityala-ion: That is a good idea. Will do.

check_tag.py
chityala-ion: I initially wanted to get the three numbers and compare them. Finally settled down to comparing the string. I will change it.

PR #74, opened 2019-12-09
TRKR-732 Allow old aperture formats so can still load pre-existing data.

rfinck:
I did an inventory of existing data and this should cover making sure we can still load it with the newest tiff reader.

### DIFF
--- a/mibidata/mibi_image.py
+++ b/mibidata/mibi_image.py
@@ -237,6 +237,10 @@ class MibiImage():
     def aperture(self, value):
         if value is not None and value not in APERTURE_MAP.values():
             try:
+                print(value, {
+                    **_DEPRECATED_APERTURE_MAP,
+                    **APERTURE_MAP
+                })
                 self._aperture = {
                     **_DEPRECATED_APERTURE_MAP,
                     **APERTURE_MAP
@@ -248,12 +252,11 @@ class MibiImage():
                                           _DEPRECATED_APERTURE_MAP[value],
                                           APERTURE_MAP)
                 )
-            except ValueError:
+            except KeyError:
                 raise ValueError(
                     'Invalid aperture code \'{}\', must use values'
                     'from the following map: {}'.format(value, APERTURE_MAP)
                 )
-        self._aperture = value
 
     @property
     def channels(self):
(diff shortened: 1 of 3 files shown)


mibidata/mibi_image.py
```diff
@@ -229,10 +236,20 @@ def aperture(self):
     @aperture.setter
     def aperture(self, value):
         if value is not None and value not in APERTURE_MAP.values():
-            raise ValueError(
-                'Invalid aperture code \'{}\', must use values'
-                'from the following map: {}'.format(value, APERTURE_MAP)
-            )
+            if value in _DEPRECATED_APERTURE_MAP:
```
rileysparsons: Great. In the future we can look to simplify `libted` pipeline app by removing this same logic from the scripts there.

mibidata/tests/test_mibi_image.py
```diff
@@ -138,6 +138,18 @@ def test_check_folder_without_fov_id(self):
         self.assertEqual(image.fov_id, 'Point1')
         self.assertEqual(image.folder, 'Point1/RowNumber0/Depth_Profile0')
 
+    def test_convert_from_deprecated_aperture(self):
+        with warnings.catch_warnings(record=True) as w:
+            image = mi.MibiImage(TEST_DATA, TUPLE_LABELS,
```
rileysparsons: 👍

setup.py
```diff
@@ -7,7 +7,7 @@
 setup(name='mibitracker-client',
       author='IONpath, Inc.',
       author_email='mibitracker-support@ionpath.com',
-      version='1.2.0',
+      version='1.2.3',
```
rileysparsons: Thanks, missed this myself on the last PR.

mibidata/mibi_image.py
```diff
@@ -229,10 +236,20 @@ def aperture(self):
...
+                warnings.warn(
+                    'Deprecated aperture code \'{}\', converting to \'{}\'. In '
+                    'a future version, values from the following map will be '
+                    'required: {}'.format(value,
+                                          _DEPRECATED_APERTURE_MAP[value],
+                                          APERTURE_MAP)
+                )
+                value = _DEPRECATED_APERTURE_MAP[value]
```
rileysparsons: @rfinck - I just realized that you're performing the mapping in the new setter. This is actually different behavior than what was happening before. All I did was check if the passed parameter for aperture was a valid code (e.g. 'B' or 'C')

rileysparsons:
Tested and works. Let's merge and tag `v1.2.3`

PR #75, opened 2019-12-18
DPPA-646: Update tiff info method to validate aperture

chris-brent:


jaytarolli:
For my own sake as I go through and review this, can you explain why we're mapping apertures in the first place? Is the expectation that TIFFs downloaded from MIBItracker will have the letter code stored in the metadata and when loaded with this reader converted to a string representation? Or is it the other way around?

mibidata/mibi_image.py
```diff
@@ -235,28 +235,7 @@ def aperture(self):
...
-                    'required: {}'.format(value,
-                                          _DEPRECATED_APERTURE_MAP[value],
-                                          APERTURE_MAP)
-                )
-            except KeyError:
-                raise ValueError(
-                    'Invalid aperture code \'{}\', must use values'
-                    'from the following map: {}'.format(value, APERTURE_MAP)
-                )
+        self._aperture = MibiImage.validate_aperture(value)
```
rfinck: Why not `self.validate_aperture`

mibidata/tiff.py
```diff
@@ -345,6 +344,9 @@ def _convert_from_previous(description):
         warnings.warn(
             'The "folder" attribute is required if "fov_id" is specified. '
             'Setting "folder" to {}.'.format(description['mibi.folder']))
+    if description.get('mibi.aperture'):
+        description['mibi.aperture'] = mi.MibiImage.validate_aperture(
+            description['mibi.aperture'])
```
rfinck: This is much nicer than what I had suggested.

rfinck:
@jaytarolli 

> For my own sake as I go through and review this, can you explain why we're mapping apertures in the first place? Is the expectation that TIFFs downloaded from MIBItracker will have the letter code stored in the metadata and when loaded with this reader converted to a string representation? Or is it the other way around?

TIFFs created by the MIBIscope software now use the aperture code, which is the same code that MIBItracker users. This mapping is for backwards compatibility, so that any old ones that contain the string get mapped _to_ the right code. The early access software left the aperture blank in the TIFFs, but some custom ones of ours wrote these strings in and we want them to still work.

jaytarolli:
@rfinck Got it, thanks. I'll update that in the specification doc as well.

mibidata/mibi_image.py
chris-brent: Good point, I also didn't like the name so I changed it.

mibidata/tiff.py
chris-brent: 👍

mibidata/mibi_image.py
```diff
@@ -307,7 +307,14 @@ def __str__(self):
         return s
 
     @staticmethod
-    def validate_aperture(value):
+    def parse_aperture(value):
+        """
+        Args:
+            value: A string representing an aperture code.
+        Returns:
+            an aperture code (e.g. 'A' or 'B') matching aperture width
```
rfinck: **A**n aperture code

mibidata/mibi_image.py
```diff
@@ -307,7 +307,14 @@ def __str__(self):
...
+            used during image acquisition.
```
rfinck: Please include a `Raises` section in the docstring.

chris-brent:
@rileysparsons Here's my PR

mibidata/mibi_image.py
```diff
@@ -235,28 +235,7 @@ def aperture(self):
...
-                    'required: {}'.format(value,
-                                          _DEPRECATED_APERTURE_MAP[value],
-                                          APERTURE_MAP)
-                )
-            except KeyError:
-                raise ValueError(
-                    'Invalid aperture code \'{}\', must use values'
-                    'from the following map: {}'.format(value, APERTURE_MAP)
-                )
+        self._aperture = self.parse_aperture(value)
```
rileysparsons: Good stuff!

mibidata/mibi_image.py
chris-brent: Done.

PR #76, opened 2020-01-04
DPPA-645: Check required metadata fields when saving MIBItiff

emma-munch:
This pull request is for changes to tiff.py in the mibitracker-client repo. The write() function now raises errors if fov_id, fov_name, run, folder, dwell, scans, mass_gain, mass_offset, or time_resolution metadata fields are missing (in addition to coordinates and size which already raised errors). 

I do have a couple quick questions indirectly related to this ticket: 

- In write(), an error is raised if the image.targets metadata field is left blank, but the MIBItiff Specifications google doc says that this field is optional. Should this be changed in write()?

- If tiff.py the metadata is for the channel is named image.masses and image.targets, but in the google doc it denotes them as channel.mass and channel.target? Should the names in the doc be plural to match the actual names in the code? (this could not even be referring to same attributes/it is super trivial but just thought I would ask)

### DIFF
--- a/mibidata/tiff.py
+++ b/mibidata/tiff.py
@@ -94,15 +94,16 @@ def write(filename, image, sed=None, optical=None, ranges=None,
         raise ValueError('image must be a mibidata.mibi_image.MibiImage '
                          'instance.')
     missing_required_metadata = [m for m in REQUIRED_METADATA_ATTRIBUTES
-                                 if not getattr(image, meta_attr)]
+                                 if not getattr(image, m)]
     if missing_required_metadata:
         if len(missing_required_metadata) == 1:
             missing_metadata_error = (f'{missing_required_metadata[0]} is '
                                       f'required and may not be None.')
-        else:    
-            missing_metadata_error = (f'{', '.join(missing_required_metadata)}'
+        else:
+            missing_metadata_error = (f'{", ".join(missing_required_metadata)}'
                                       f' are required and may not be None.')
         raise ValueError(missing_metadata_error)
+
     if write_float is not None:
         raise ValueError('`write_float` has been deprecated. Please use the '
                          '`dtype` argument instead.')
--- a/mibidata/tiff.py
+++ b/mibidata/tiff.py
@@ -31,9 +31,9 @@ _MAX_DENOMINATOR = 1000000
 # Encoding of tiff tags.
 ENCODING = 'utf-8'
 
-REQUIRED_METADATA_ATTRIBUTES = ('fov_id', 'fov_name', 'run', 'folder',
-                                'dwell', 'scans', 'mass_gain', 'mass_offset',
-                                'time_resolution', 'coordinates')
+REQUIRED_METADATA_ATTRIBUTES = ('fov_id', 'fov_name', 'run', 'folder', 'dwell',
+                                'scans', 'mass_gain', 'mass_offset', 'time_resolution',
+                                'coordinates', 'size', 'masses', 'targets')
 
 def _micron_to_cm(arg):
     """Converts microns (1cm = 1e4 microns) to a fraction tuple in cm."""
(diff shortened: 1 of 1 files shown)


jaytarolli:
> If tiff.py the metadata is for the channel is named image.masses and image.targets, but in the google doc it denotes them as channel.mass and channel.target? Should the names in the doc be plural to match the actual names in the code? (this could not even be referring to same attributes/it is super trivial but just thought I would ask)

In the MibiImage class, these are stored as lists, but when saved to a MIBItiff file, the mass and target are saved per channel as opposed to a list of each. That way, each page of the TIFF file contains the target and mass for corresponding to that image data. Every other piece of metadata is saved per image though, that's why the prefixes are different.

> In write(), an error is raised if the image.targets metadata field is left blank, but the MIBItiff Specifications google doc says that this field is optional. Should this be changed in write()?

I'm inclined to say both masses and targets should be required, as the current validation enforces. Unless @rfinck has a reason otherwise, let's mark both of those as required in the TIFF specification document as well.

mibidata/tiff.py
```diff
@@ -31,6 +31,9 @@
 # Encoding of tiff tags.
 ENCODING = 'utf-8'
 
+REQUIRED_METADATA_ATTRIBUTES = ('fov_id', 'fov_name', 'run', 'folder',
```
jaytarolli: As it stands, this isn't used anywhere, though see my next comment before deciding to remove.

mibidata/tiff.py
```diff
@@ -90,6 +94,24 @@ def write(filename, image, sed=None, optical=None, ranges=None,
                          'instance.')
     if image.coordinates is None or image.size is None:
         raise ValueError('Image coordinates and size must not be None.')
+    if image.fov_id is None:
```
jaytarolli: Since you've already added `REQUIRED_METADATA_ATTRIBUTES` above, I think you could simplify all of this by looping through each and checking if None with `getattr()`. In that case, it's fine to print out an error message containing the attribute name instead of a more readable version, e.g. `fov_name is required and may not be None`.

It might be even better to collect a list of all of the attributes that are None and concat the list to print a single error message. That way a user doesn't have to go through and set each one before finding out the next one is required, having to get 10 ValueErrors before succeeding. For example: `fov_name, run, and folder are required and may not be None`.

emma-munch: > Since you've already added REQUIRED_METADATA_ATTRIBUTES above, I think you could simplify all of this by looping through each and checking if None with` getattr()`
I now use the list of required attributes and loop through

> It might be even better to collect a list of all of the attributes that are None and concat the list to print a single error message.
fixed so error message lists all missing metadata fields

mibidata/tiff.py
```diff
@@ -88,10 +93,24 @@ def write(filename, image, sed=None, optical=None, ranges=None,
     if not isinstance(image, mi.MibiImage):
         raise ValueError('image must be a mibidata.mibi_image.MibiImage '
                          'instance.')
-    if image.coordinates is None or image.size is None:
-        raise ValueError('Image coordinates and size must not be None.')
-    if image.masses is None or image.targets is None:
-        raise ValueError('Image channels must contain both masses and targets.')
+    missing_required_metadata = []
+    for meta_attr in REQUIRED_METADATA_ATTRIBUTES:
```
jaytarolli: I think you could do `missing_required_metadata = [m for m in REQUIRED_METADATA_ATTRIBUTES if not getattr(image, m)]`.

mibidata/tiff.py
```diff
@@ -88,10 +93,24 @@ def write(filename, image, sed=None, optical=None, ranges=None,
...
+        if len(missing_required_metadata) == 1:
+            missing_metadata_error = (f'{missing_required_metadata[0]} is '
+                                      f'required and may not be None.')
+        else:
+            missing_fields = ''
+            for i, meta_attr in enumerate(missing_required_metadata):
+                if i == len(missing_required_metadata) - 1:
+                    missing_fields += f'{meta_attr} '
+                else:
+                    missing_fields += f'{meta_attr}, '
```
jaytarolli: Instead of looping through here, you can use `', '.join(missing_required_metadata)`

emma-munch:
@jaytarolli I've made all requested changes so take a look when you have the chance.

PR #77, opened 2020-01-07
DPPA-579: Avoid repetition between tiff.info and tiff.read

emma-munch:
Added very lightweight wrapper function for 2 main lines of shared code between the tiff.info and tiff.read called _get_page data() which updates the metadata for the tiff and list of channels.

Updated helper function _get_description() to return the image type as well as description to minimize redundancy

### DIFF
--- a/mibidata/mibi_image.py
+++ b/mibidata/mibi_image.py
@@ -205,14 +205,8 @@ class MibiImage():
     def folder(self, value):
         """Enforce consistency with fov_id."""
         if value:
-            fov = value.split('/')[0]
-            if not self.fov_id:
-                self._fov_id = fov
-            elif self.fov_id != fov:
-                raise ValueError('fov_id must match folder, but here '
-                                 'fov_id={} and you are trying to set folder '
-                                 'to {}.'.format(self.fov_id, value))
             self._folder = value
+            self.match_fov_folder(value, self._fov_id, 'folder', 'fov_id')
 
     @property
     def fov_id(self):
@@ -221,13 +215,8 @@ class MibiImage():
     @fov_id.setter
     def fov_id(self, value):
         """Enforce consistency with folder."""
-        if not self.folder:
-            self._folder = value
-        elif value != self.folder.split('/')[0]:
-            raise ValueError('fov_id must match folder, but here '
-                             'folder={} and you are trying to set fov_id '
-                             'to {}.'.format(self.folder, value))
         self._fov_id = value
+        self.match_fov_folder(value, self._folder, 'fov_id', 'folder')
 
     @property
     def aperture(self):
@@ -341,6 +330,28 @@ class MibiImage():
                     'from the following map: {}'.format(value, APERTURE_MAP)
                 )
         return aperture
+
+
+    @staticmethod
+    def match_fov_folder(value, field, value_name, field_name):
+    """Enforces consistency between fov_id and folder.
+
+    Args:
+        value: Value to match with.
+        field: Field that needs to be updated to value.
+        value_name: String representing whether matching folder or fov_id field.
+        field_name: String representing whether updating folder or fov_id field.
+    """
+        if field_name == 'fov_id':
+            value = value.split('/')[0]
+        if not field:
+            warnings.warn('The "{field_name}" attribute is required if "{value_name}" is specified. '
+                      'Setting "{field_name}" to {value}.')
+        elif field != value:
+            warnings.warn(
+            'The attribute "{field_name}" must match "{value_name}". '
+            'Changing "{field_name}" from {} to {}.'.format(field, value))
+        field = value
 
     def metadata(self):
         """Returns a dictionary of the image's metadata."""
--- a/mibidata/mibi_image.py
+++ b/mibidata/mibi_image.py
@@ -334,23 +334,26 @@ class MibiImage():
 
     @staticmethod
     def match_fov_folder(value, field, value_name, field_name):
-    """Enforces consistency between fov_id and folder.
+        """Enforces consistency between fov_id and folder.
 
-    Args:
-        value: Value to match with.
-        field: Field that needs to be updated to value.
-        value_name: String representing whether matching folder or fov_id field.
-        field_name: String representing whether updating folder or fov_id field.
-    """
+        Args:
+            value: Value to match with.
+            field: Field that needs to be updated to value.
+            value_name: String representing whether matching folder
+                        or fov_id field.
+            field_name: String representing whether updating folder
+                        or fov_id field.
+        """
         if field_name == 'fov_id':
             value = value.split('/')[0]
         if not field:
-            warnings.warn('The "{field_name}" attribute is required if "{value_name}" is specified. '
-                      'Setting "{field_name}" to {value}.')
+            warnings.warn(
+                f'The "{field_name}" attribute is required if "{value_name}" '
+                f'is specified. Setting "{field_name}" to {value}.')
         elif field != value:
             warnings.warn(
-            'The attribute "{field_name}" must match "{value_name}". '
-            'Changing "{field_name}" from {} to {}.'.format(field, value))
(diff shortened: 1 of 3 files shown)


rfinck:
I think it'd be nice as part of this to get rid of the code duplication in `_convert_from_previous` that's also in the MibiImage class, probably by using a static method there similar to in https://github.com/ionpath/mibitracker-client/pull/75/files

jaytarolli:
I think there probably is still a bit of repeated code between `tiff.read()` and `tiff.info()` that could be refactored. It seems to me that `tiff.info()` could be replaced with:

```
def info(filename):
    im = tiff.read(filename)
    return im.metadata()
```

I don't see anywhere in mibitracker-client or mibitof in which `tiff.info()` is called, so is it even needed?

rfinck:
> I don't see anywhere in mibitracker-client or mibitof in which tiff.info() is called, so is it even needed?

Yes, it's much more efficient to just read the metadata when you don't need the data. We use it in mibitracker and elsewhere for that.

jaytarolli:
Alright, then ignore my suggestion to have `tiff.info()` call `tiff.read()`.

emma-munch:
6844982: Added static method MibiImage.match_fov_folder() that warns the user that the fov_id and folder fields need to match if they do not and returns value to update with. Called in fov_id and folder setter methods as well as tiff._convert_from_previous() method whenever metadata for fov_id or folder are missing.

9849500: Bug was that was missing if statement in _convert_from_previous(), so always called match_fov_folder() and warned unecessarily. Fixed in this commit.

3534b62: The `match_fov_folder()` method now returns the value in order to ensure `description` in `_convert_from_previous()` actually gets updated to matched value. Necessary, because when changing `description` dictionary, values are immutable strings so cannot be changed via pass-by-reference parameter.

mibidata/mibi_image.py
```diff
@@ -342,6 +333,34 @@ def parse_aperture(value):
                 )
         return aperture
 
+
+    @staticmethod
+    def match_fov_folder(value, field, value_name='valuename',
```
jaytarolli: The parameters `value_name` and `field_name` seem to be required and the default values specified in the signature are never used. Can you just leave off the defaults? Or should they default to one combination of `value_name` and `field_name`?

mibidata/tiff.py
```diff
@@ -343,26 +342,40 @@ def _convert_from_previous(description):
...
+        field = 'mibi.fov_id'
+        matchNeeded = True
+    if description.get('mibi.fov_id'):
+        description['mibi.folder'] = ''
+        value = 'mibi.fov_id'
+        field = 'mibi.folder'
+        matchNeeded = True
+    if matchNeeded:
+        description[field] = mi.MibiImage.match_fov_folder(description[value],
+                                description[field], value[5:], field[5:])
```
jaytarolli: I would suggest filling `field_name` and `value_name` by doing `value.split('.')[1]` rather than assume the length of the prefix. Though, can you just set `value` and `field` above to `folder` or `fov_id` without the mibi prefix?

mibidata/tiff.py
```diff
@@ -343,26 +342,40 @@ def _convert_from_previous(description):
...
-                description['mibi.fov_id']))
-    if (not description.get('mibi.folder') and description.get('mibi.fov_id')
-            and description.get('mibi.fov_id').startswith('FOV')):
-        description['mibi.folder'] = description['mibi.fov_id']
-        warnings.warn(
-            'The "folder" attribute is required if "fov_id" is specified. '
-            'Setting "folder" to {}.'.format(description['mibi.folder']))
+    if description.get('mibi.folder'):
+        description['mibi.fov_id'] = ''
+        value = 'mibi.folder'
```
jaytarolli: It might be cleaner to just call `mi.MibiImage.match_fov_folder()` in each of these `if` statements. Or can both `if` statements evaluate to True and you need to override the values when `mibi.folder` is present with those for `mibi.fov_id`?

emma-munch: I'm not completely sure if there is a case where fov_id and folder are both present in the description but don't match. I was just trying to be on the safe side if somehow the old JSON descriptions didn't enforce this and both if branches execute.

mibidata/tiff.py
```diff
@@ -343,26 +342,44 @@ def _convert_from_previous(description):
...
+        field = 'mibi.fov_id'
+        value_name = 'folder'
+        field_name = 'fov_id'
+        matchNeeded = True
+    if description.get('mibi.fov_id'):
+        description['mibi.folder'] = ''
+        value = 'mibi.fov_id'
+        field = 'mibi.folder'
+        value_name = 'fov_id'
```
jaytarolli: > I would suggest filling field_name and value_name by doing value.split('.')[1] rather than assume the length of the prefix. Though, can you just set value and field above to folder or fov_id without the mibi prefix?

Sorry, I see now that `value` and `field` need to have the `'mibi.'` prefix, but rather than define new variables here, below when you call `mi.MibiImage.match_fov_folder()` use `value.split('.')[1]`.

mibidata/tiff.py
```diff
@@ -343,26 +342,41 @@ def _convert_from_previous(description):
...
+        description['mibi.fov_id'] = ''
+        value = 'mibi.folder'
+        field = 'mibi.fov_id'
+        matchNeeded = True
+    if description.get('mibi.fov_id'):
+        description['mibi.folder'] = ''
+        value = 'mibi.fov_id'
+        field = 'mibi.folder'
+        matchNeeded = True
+    if matchNeeded:
```
jaytarolli: If somehow neither fov_id nor folder are in the metadata, `matchNeeded` won't have been define when you check it here, which will cause an error.

Also, I should have caught earlier that camelCase is apparently shunned in Python, and so this should be `match_needed`.

mibidata/mibi_image.py
```diff
@@ -342,6 +333,34 @@ def parse_aperture(value):
...
+            field_name: String representing whether updating folder
+                        or fov_id field.
+        Returns:
+            value: Value to set field to.
+        """
+        if field_name == 'fov_id' and value_name == 'folder':
+            value = value.split('/')[0]
+        if not field:
+            warnings.warn(
+                f'The "{field_name}" attribute is required if "{value_name}" '
```
chris-brent: parameter not attribute

mibidata/mibi_image.py
```diff
@@ -342,6 +333,34 @@ def parse_aperture(value):
                 )
         return aperture
 
+
+    @staticmethod
+    def match_fov_folder(value, field, value_name,
```
chris-brent: Nit pick comment, but shouldn't this be match_field or something similar? If we're passing in a generic parameter names then should the function name also be generic? I also prefer validate to match in cases like that, but Python might disagree.

mibidata/mibi_image.py
```diff
@@ -342,6 +333,34 @@ def parse_aperture(value):
...
+        """
+        if field_name == 'fov_id' and value_name == 'folder':
+            value = value.split('/')[0]
+        if not field:
+            warnings.warn(
+                f'The "{field_name}" attribute is required if "{value_name}" '
+                f'is specified. Setting "{field_name}" to {value}.')
+        elif field != value:
+            warnings.warn(
+                f'The attribute "{field_name}" must match "{value_name}". '
```
chris-brent: parameter not attribute

mibidata/mibi_image.py
chris-brent: This feels more like an exception than a warning. Do we ever want a user to be able to set folder? Should we just always derive it from fov_id? @jaytarolli what are you using in MIBIO for the -Filtered images?

jaytarolli: When an isobaric corrected or filtered image is saved, the default behavior of MIBI/O is to append a (user-customizable) string to the end of the `fov_id`, e.g.: fov-1-scan-1 - Filtered. `folder` is then set by replacing first part of the folder value with the `fov_id`.

PR #78, opened 2020-01-08
DPPA 578: Add methods to allow the user to add optional metadata attributes to MibiImage

emma-munch:
Added add_attr(kwargs) and remove_attr(attributes) to MibiImage class.
add_attr():
- if user tries to add an attribute that already exists, will just update the value of attr
- allows user to specify user-defined and required attributes
- able to parse the format of data attribute accordingly
- if user defined, adds to _user_defined_attributes list

remove_attr():
- if user tries to remove required attribute, just sets it to None, otherwise removes attribute from instance and _user_defined_attributes list
- if user tries to remove attribute that does exists, just moves on
- warns user that they tried to delete required attributes
- warns user that tried to remove nonexistent attributes

Let me know if all of this functionality is ok.

### DIFF
--- a/mibidata/mibi_image.py
+++ b/mibidata/mibi_image.py
@@ -225,11 +225,11 @@ class MibiImage():
             else:
                 _no_attr.append(attr)
         if _required_rem:
-            raise ValueError(f'{", ".join(_required_rem)} are required and '
-                             f'cannot be removed.')
+            warnings.warn(f'{", ".join(_required_rem)} are required and '
+                          f'were not removed.')
         if _no_attr:
-            raise ValueError(f'{", ".join(_no_attr)} are not attributes of '
-                             f'this instance.')
+            warnings.warn(f'{", ".join(_no_attr)} are not attributes of '
+                          f'this instance.')
 
 
     @property
--- a/mibidata/mibi_image.py
+++ b/mibidata/mibi_image.py
@@ -222,11 +222,12 @@ class MibiImage():
                 setattr(self, attr, None)
             elif hasattr(self, attr):
                 delattr(self, attr)
+                self._user_defined_attributes.remove(attr)
             else:
                 _no_attr.append(attr)
         if _required_rem:
             warnings.warn(f'{", ".join(_required_rem)} are required and '
-                          f'were not removed.')
+                          f'were not removed. Attributes were set to None.')
         if _no_attr:
             warnings.warn(f'{", ".join(_no_attr)} are not attributes of '
                           f'this instance.')
--- a/mibidata/mibi_image.py
+++ b/mibidata/mibi_image.py
@@ -213,11 +213,12 @@ class MibiImage():
             setattr(self, key, value)
 
 
-    def remove_attr(self, attributes):
+    def remove_attr(self, *args):
         """Removes user-defined attributes from the class instance in use.
 
         Args:
-            attributes: A list of user-defined attributes for deletion.
+            attributes: A single string or a list of user-defined attributes
+                        for deletion.
 
         Raises:
             ValueError: Raised if
@@ -225,16 +226,16 @@ class MibiImage():
                 * attempts to remove a required attribute.
                 * an attribute is not defined for this instance.
         """
-        required_rem = [attr for attr in attributes if attr
+        required_rem = [attr for attr in args if attr
                         in SPECIFIED_METADATA_ATTRIBUTES]
-        no_attr = [attr for attr in attributes if not hasattr(self, attr)]
+        no_attr = [attr for attr in args if not hasattr(self, attr)]
         if required_rem:
             if len(required_rem) == 1:
                 required_error = (f'{required_rem[0]} is a required '
-                                  f'attribute and was not removed.')
+                                  f'attribute.')
             else:
                 required_error = (f'{", ".join(required_rem)} are required '
-                                  f'attributes and were not removed.')
+                                  f'attributes.')
             raise ValueError(required_error)
         if no_attr:
             if len(no_attr) == 1:
@@ -244,7 +245,7 @@ class MibiImage():
                 required_error = (f'{", ".join(no_attr)} are not attributes '
                                   f'of this instance.')
             raise ValueError(required_error)
-        for attr in attributes:
+        for attr in args:
             delattr(self, attr)
             self._user_defined_attributes.remove(attr)
 
--- a/mibidata/mibi_image.py
+++ b/mibidata/mibi_image.py
@@ -199,21 +199,19 @@ class MibiImage():
             defined for this instance.
 
         """
-        already_defined = []
-        for key, value in kwargs.items():
-            if not hasattr(self, key):
-                self._user_defined_attributes.append(key)
-                setattr(self, key, value)
-            else:
-                already_defined.append(key)
+        already_defined = [attr for attr in kwargs if hasattr(self, attr)]
         if already_defined:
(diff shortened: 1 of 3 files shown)


mibidata/mibi_image.py
```diff
@@ -185,6 +185,54 @@ def __init__(self, data, channels, **kwargs):
...
+                    values.
+        """
+        date = kwargs.pop('date', self.date)
+        datetime_format = kwargs.pop('datetime_format', _DATETIME_FORMAT)
+        try:
+            self.date = datetime.datetime.strptime(date, datetime_format)
+        except TypeError:  # Given as datetime obj already, or None.
+            self.date = date
+        for key, value in kwargs.items():
+            if key not in SPECIFIED_METADATA_ATTRIBUTES:
```
jaytarolli: My opinion is that in an "add attribute" function, if the attribute already exists, an error should be thrown. That, to me, seems more intuitive. I would, before setting anything, check the keys in `kwargs` and if they are already in the metadata, raise an error.

mibidata/mibi_image.py
```diff
@@ -185,6 +185,54 @@ def __init__(self, data, channels, **kwargs):
...
+        _no_attr = []
+        for attr in attributes:
+            if attr in _required_attr:
+                setattr(self, attr, None)
+            elif hasattr(self, attr):
+                delattr(self, attr)
+                self._user_defined_attributes.remove(attr)
+            else:
+                _no_attr.append(attr)
+        if _required_rem:
```
jaytarolli: Similarly, here, I would not warn but rather raise an error if a required attribute is being removed.

mibidata/mibi_image.py
```diff
@@ -185,6 +185,54 @@ def __init__(self, data, channels, **kwargs):
...
+            warnings.warn(f'{", ".join(_required_rem)} are required and '
+                          f'were not removed. Attributes were set to None.')
+        if _no_attr:
```
jaytarolli: And finally here, I would raise an error as well.

mibidata/mibi_image.py
```diff
@@ -185,6 +185,54 @@ def __init__(self, data, channels, **kwargs):
...
+            warnings.warn(f'{", ".join(_no_attr)} are not attributes of '
```
jaytarolli: When these raise errors, I like what you did in DPPA-645 to use "is" or "are" based whether there is one error to report on or multiple.

mibidata/mibi_image.py
```diff
@@ -185,6 +185,77 @@ def __init__(self, data, channels, **kwargs):
...
+            kwargs: A mapping of arguments for a user to add multiple
+                    attributes with their respecitve
+                    values.
+        Raises:
+            ValueError: Raised if attempts to set an attribute that is already
+            defined for this instance.
+
+        """
+        already_defined = []
+        for key, value in kwargs.items():
```
jaytarolli: Let's do check in the loop without setting the values, then raise the error if one or more attributes already exist. That way, when the user gets the error, he/she can assume the operation failed without some of the attributes having the desired value. Then, after the `if already_defined:` check passes, you can loop again and set the attribute values.

mibidata/mibi_image.py
```diff
@@ -185,6 +185,77 @@ def __init__(self, data, channels, **kwargs):
...
+            if not hasattr(self, key):
+                self._user_defined_attributes.append(key)
+                setattr(self, key, value)
+            else:
+                already_defined.append(key)
+        if already_defined:
+            if len(already_defined) == 1:
+                already_defined_error = (f'{already_defined[0]} is '
```
jaytarolli: 👍

mibidata/mibi_image.py
```diff
@@ -185,6 +185,77 @@ def __init__(self, data, channels, **kwargs):
...
+        Raises:
+            ValueError: Raised if
+
+                * attempts to remove a required attribute.
+                * an attribute is not defined for this instance.
+        """
+        _required_attr = SPECIFIED_METADATA_ATTRIBUTES
+        _required_rem = []
+        _no_attr = []
+        for attr in attributes:
```
jaytarolli: Same thing about raising the errors before actually removing any attributes.

mibidata/mibi_image.py
```diff
@@ -185,6 +185,70 @@ def __init__(self, data, channels, **kwargs):
...
+        for key, value in kwargs.items():
+            self._user_defined_attributes.append(key)
+            setattr(self, key, value)
+
+
+    def remove_attr(self, attributes):
+        """Removes user-defined attributes from the class instance in use.
+
+        Args:
+            attributes: A list of user-defined attributes for deletion.
```
jaytarolli: The only suggestion left I would make is allow attributes to be a string or a list of strings so that if a user is trying to remove a single attribute he/she doesn't have to pass it as a list of one item.

mibidata/mibi_image.py
```diff
@@ -185,6 +185,70 @@ def __init__(self, data, channels, **kwargs):
...
+                * attempts to remove a required attribute.
+                * an attribute is not defined for this instance.
+        """
+        required_rem = [attr for attr in attributes if attr
+                        in SPECIFIED_METADATA_ATTRIBUTES]
+        no_attr = [attr for attr in attributes if not hasattr(self, attr)]
+        if required_rem:
+            if len(required_rem) == 1:
+                required_error = (f'{required_rem[0]} is a required '
+                                  f'attribute and was not removed.')
```
jaytarolli: I think you can take out the `'and was not removed'` part since now if an error is raised nothing was actually removed.

jaytarolli:
@emma-munch I tried this out and `remove_attr` worked when I passed a single string, but when I passed a list of strings, I got:

```
In [10]: a.remove_attr(['temp1', 'temp2'])
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-10-e8d3207736cf> in <module>
----> 1 a.remove_attr(['temp1', 'temp2'])

~\source\repos\mibitracker-client\mibidata\mibi_image.py in remove_attr(self, *args)
    229         required_rem = [attr for attr in args if attr
    230                         in SPECIFIED_METADATA_ATTRIBUTES]
--> 231         no_attr = [attr for attr in args if not hasattr(self, attr)]
    232         if required_rem:
    233             if len(required_rem) == 1:

~\source\repos\mibitracker-client\mibidata\mibi_image.py in <listcomp>(.0)
    229         required_rem = [attr for attr in args if attr
    230                         in SPECIFIED_METADATA_ATTRIBUTES]
--> 231         no_attr = [attr for attr in args if not hasattr(self, attr)]
    232         if required_rem:
    233             if len(required_rem) == 1:

TypeError: hasattr(): attribute name must be string
```

emma-munch:
@jaytarolli review when you can, no more asterisks for remove_attr()

mibidata/mibi_image.py
```diff
@@ -226,9 +226,11 @@ def remove_attr(self, *args):
                 * attempts to remove a required attribute.
                 * an attribute is not defined for this instance.
         """
-        required_rem = [attr for attr in args if attr
+        if isinstance(attributes, str):
+            attributes = list(attributes)
```
jaytarolli: You need to do `attributes = [attributes]` otherwise it just creates a list of characters in the string passed.

emma-munch:
sorry this review was so dragged out! thanks for all your patience.

jaytarolli:
This looks good to me now!

PR #80, opened 2020-01-31
DPPA-683: Add method to read selected channels of TIFF

emma-munch:


mibidata/tests/test_tiff.py
```diff
@@ -105,6 +105,22 @@ def test_sims_only(self):
         self.assertIsNone(label)
         self.assertEqual(image.data.dtype, np.float32)
 
+    def test_sims_selected_channels(self):
+        tiff.write(self.filename, self.float_image)
+        image = tiff.read(self.filename, inc_channels=CHANNELS[1:3])
+        print(image.channels)
```
jaytarolli: Remove print statement

mibidata/tiff.py
```diff
@@ -361,39 +371,59 @@ def _convert_from_previous(description):
...
     Returns:
         A dictionary of metadata as could be supplied as kwargs to
-        :class:`mibidata.mibi_image.MibiImage`, except with a ``channels`` key
-        whose value is a list of (mass, target) tuples.
+        :class:`mibidata.mibi_image.MibiImage`, except with a ``conjugates`` key
+        whose value is a list of (mass, target) tuples. If inc_channels is set,
+        then the metadata ``conjugates`` key only contains information for those
+        channels.
     """
+    if (inc_channels and not isinstance(inc_channels, (list, tuple)) and not
```
jaytarolli: Since I just reviewed DPPA-466, this code seems to be very similar to a couple other instances in that other PR. Once you merge DPPA-466 and bring the changes into here, would it make sense to refactor some of this code into helper functions like `check_is_list_of_tuples` or something?

mibidata/tiff.py
```diff
@@ -361,39 +371,59 @@ def _convert_from_previous(description):
...
+
+    Returns:
+        include: Boolean representing whether channel should be included or not.
     """
-    channels.append((description['channel.mass'],
-                     description['channel.target']))
     # Get metadata on first SIMS page only
     if not metadata:
         metadata.update(_page_metadata(page, description))
+    if inc_channels:
```
jaytarolli: You could probably combine these two if statements into a single one with `and`.

mibidata/tiff.py
```diff
@@ -361,39 +371,59 @@ def _convert_from_previous(description):
...
 
     Returns:
         A dictionary of metadata as could be supplied as kwargs to
-        :class:`mibidata.mibi_image.MibiImage`, except with a ``channels`` key
-        whose value is a list of (mass, target) tuples.
+        :class:`mibidata.mibi_image.MibiImage`, except with a ``conjugates`` key
+        whose value is a list of (mass, target) tuples. If inc_channels is set,
+        then the metadata ``conjugates`` key only contains information for those
+        channels.
     """
```
jaytarolli: You'll need to include an `Raises` section now that you're checking the `inc_channels` parameter.

mibidata/tiff.py
```diff
@@ -241,14 +242,17 @@ def read(file, sims=True, sed=False, optical=False, label=False):
...
         specified). The SIMS data will be returned as a
         :class:`mibidata.mibi_image.MibiImage` instance; the other image
         types will be returned as numpy arrays. If an image type is selected to
         be returned  but is not present in the image, it will be returned as
-        None.
+        None. If returning SIMS data and the inc_channels parameter is set, only
+        those channels will be included in the MibiImage instance.
 
     Raises:
         ValueError: Raised if the input file is not of the IonpathMIBI
```
jaytarolli: You'll need to add to the `Raises` section here too.

mibidata/tests/test_tiff.py
```diff
@@ -105,6 +105,22 @@ def test_sims_only(self):
...
+        expected_image = tiff.read(self.filename)
+        expected_image = expected_image.slice_image(CHANNELS[1:3])
+        self.assertEqual(expected_image.channels, image.channels)
+        self.assertEqual(expected_image.metadata(), image.metadata())
+        self.assertEqual(expected_image, image)
```
jaytarolli: I think the asserts for `channels` and `metadata` are redudant because `MibiImage.__eq__` already includes both of those properties in it's equality check.

mibidata/tiff.py
```diff
@@ -241,24 +242,36 @@ def read(file, sims=True, sed=False, optical=False, label=False):
...
-        None.
+        None. If returning SIMS data and the inc_channels parameter is set, only
+        those channels will be included in the MibiImage instance.
 
     Raises:
-        ValueError: Raised if the input file is not of the IonpathMIBI
-            format, or if no image type selected to be returned.
+        ValueError: Raised if
+
+            * The input file is not of the IonpathMIBI format, or if no image
```
jaytarolli: Let's call this the IONpath MIBItiff format

mibidata/tiff.py
```diff
@@ -241,24 +242,36 @@ def read(file, sims=True, sed=False, optical=False, label=False):
...
+              type selected to be returned.
+            * The inc_channels parameter is not a tuple or list of
+              string tuples.
     """
     return_types = collections.OrderedDict([
         ('sims', sims), ('sed', sed), ('optical', optical), ('label', label)
     ])
     if not any((val for val in return_types.values())):
         raise ValueError('At least one image type must be specified to return.')
+    if (inc_channels and not
```
jaytarolli: If I'm reading this and the docstring correctly, `inc_channels` needs to be either a list of (mass, target) tuples or a single (mass, target) tuple, correct? 1) If that's the case, you should add tests for the case where it's a single tuple because right now the only ones are for 2 channels, which tests the list part. 2) Could this be generalized to pass in a single channel string or list of channel strings? I can imagine use cases for both.

emma-munch:
@jaytarolli  I know that in the `MibiImage` documentation it says that `channels` should always be a list of string tuples or a list of strings, but tiff.write() always casts the masses attribute to an int so when read back the channels will be (int, str) and in test_tiff.py, the images are constructed with a list of tuples in the this same format. So should I incorporate the case of passing in a list of integer masses?
Also, as a side note: the `MibiImage` objects in test_mibi_image.py are constructed with `channels` in form of ('Mass1', 'Target1'), but if one attempted to call tiff.write() on this, it would throw a value error due to [the cast in tiff.write()](https://github.com/ionpath/mibitracker-client/blob/59b397fc6a39df35b8f39e9432932a1d6b2bc183/mibidata/tiff.py#L164-L167). Should the documentation be updated or should tiff.write() be modified?

jaytarolli:
My take on this is that `channels` should support either string or integer masses because I can think of use cases for both, but also our MIBItiff specification defines the required field channel mass as "The integer mass of the isotope...". Realistically, I think it should only support integer masses, but that might a discussion for day.

As for the scope of this PR, I would say let's make sure nothing is broken by using either (int, string) or (string, string) channels. If you have places where you see inconsistencies or cases that aren't tested, let's open up new tickets for those items. This way we can make sure we don't get too far into the weeds for this one PR.

emma-munch:
left failing test in because this is what brought up a lot of the issues documented in DPPA-732

rfinck:
@jaytarolli I simplified this a bit by having the user specify targets or masses. I think we can save bigger changes to channel formatting for a later issue.

mibidata/mibi_image.py
```diff
@@ -439,77 +435,6 @@ def metadata(self):
         metadata_keys.extend(self._user_defined_attributes)
         return {key: getattr(self, key) for key in metadata_keys}
 
-    @staticmethod
-    def channels_is_list_of_tuples(channels):
```
jaytarolli: I liked the idea of having these defined as functions to make it clear what each is doing. Can we either reimplement the ones needed or document the code (e.g. line 319: `# Check if channels are tuples of mass and target`?

mibidata/mibi_image.py
```diff
@@ -617,25 +542,20 @@ def append(self, image):
...
-        if set(og_channels).intersection(set(new_channels)):
+        if set(self.channels).intersection(set(image.channels)):
             raise ValueError('Images contain overlapping channels.')
-        if MibiImage.channels_is_list_of_tuples(og_channels):
-            if not MibiImage.channels_is_list_of_tuples(new_channels):
-                raise ValueError('Channels to be appended must match form of '
+        if all((isinstance(c, tuple) and len(c) == 2 for c in self.channels)):
+            if not all((isinstance(c, tuple) and len(c) == 2
+                        for c in image.channels)):
+                raise ValueError('Channels to be appended must match form as '
```
jaytarolli: You removed from the docstring that this ValueError could be raised for

mibidata/tiff.py
```diff
@@ -282,32 +278,46 @@ def read(file, sims=True, sed=False, optical=False, label=False,
...
-                    sims_data.append(page.asarray())
+            if sims and image_type == 'sims' and _include_page(
+                    description, masses, targets):
+                _get_page_data(page, description, metadata, channels)
+                sims_data.append(page.asarray())
             elif return_types.get(image_type):
                 to_return[image_type] = page.asarray()
     if sims:
-        if inc_channels and not sims_data:
+        if (targets or masses) and not sims_data:
```
jaytarolli: Does it make sense to restrict this to only masses *or* targets can be specified, in case someone specifies different lists for each? Also, what if someone passes a list of tuples of (mass, target)?

mibidata/mibi_image.py
rfinck: I'm not opposed to these but think they belong in DPPA-732 as the scope of this PR should be tiff reader only and not extend to MibiImage changes.

mibidata/mibi_image.py
rfinck: Good catch, I was over-eager in my revert.

mibidata/tiff.py
rfinck: I wondered about restricting to one or the other as well, and could be convinced to do that. I don't have a strong use-case for needing to specify a mix, the closest I got was a LOO-type experiment where you've extracted the empty channel when the target was left out, but in that case you could just specify masses across the board.

mibidata/mibi_image.py
jaytarolli: Sure, it can be done as part of DPPA-732.

mibidata/tiff.py
jaytarolli: I can't think of any use cases where specifying both would be needed, was more just concerned that the option should not be there to avoid potential issues when being used. For example, if someone accidentally specified two lists that were not the same, the data returned might not be what is expected and tough to track down why.

I do think it makes sense to be able to pass a list of tuples though, because I can image a scenario where someone uses a subset of `MibiImage.channels` to load a TIFF, in which case the list would be one of tuples.

rfinck: I updated to only allow masses _or_ targets to be specified.

I see you point about a list of tuples, but it's not hard to extra either channels or masses from that and I'm hesitant to litter the kwargs more than they already are, so can we put that off and possible revisit as part of the general channel formatting issue?

PR #82, opened 2020-01-31
DPPA-466: Mibi image append shouldn't add channels and targets if the masses and targets are not unique

emma-munch:


jaytarolli:
Looks good to me!

PR #85, opened 2020-02-12
DPPA-741: Update get_channel_data with new MIBItracker routes

jaytarolli:


jaytarolli:
@chris-brent Yes(?) because it outputs:

`HTTPError: [Errno 404 Client Error: Not Found for url: https://backend-dot-mibitracker-internal.appspot.com/images/11558/channel_url/?channel=cell_boundares] {'message': 'Channel not found'}`

from the `raise_for_status()` call. I guess I could consume that if the response is a 404 and raise a `MibiTrackerError` to be cleaner, and then let `raise_for_status` catch all the rest.

PR #86, opened 2020-02-18
DPPA-753 Add gray to composite 

rfinck:
This adds 'Gray' to the general compositing function. It also prevents a possible bug if the first channel is at min brightness when applying MIBItracker overlay settings, simplifies applying the hue to each channel.

Addresses request added post-closure to #64.


mibidata/color.py
```diff
@@ -208,7 +176,7 @@ def composite(image, color_map, gamma=1/3, min_scaling=10):
         color_map: A dictionary keyed by color with values of channel names
             corresponding to a subset of those in the MibiImage. The
             allowed colors are 'Cyan', 'Yellow', 'Magenta', 'Green',
-            'Orange', 'Violet', 'Red' and 'Blue'.
+            'Orange', 'Violet', 'Red', 'Blue' and 'Gray'.
```
jaytarolli: I'll leave my strong feelings regarding the Oxford comma out of this review.

mibidata/color.py
```diff
@@ -279,28 +254,28 @@ def compose_overlay(image, overlay_settings):
...
-        float_array = int_array / float(range_max)
-        float_array[int_array > 0] += item.get('brightness', 0)
+        range_max = item.get('intensity_higher', array.max())
+        array[array > range_max] = range_max
+        array[array < range_min] = 0
+        array = array / float(range_max)
+        array[array > 0] += item.get('brightness', 0)
         # Apply default filter even if no blurring
-        ndimage.filters.convolve(float_array, kernel, output=float_array)
+        array = ndimage.filters.convolve(array, kernel)
```
jaytarolli: Very minor, but now there is a mix of return and inplace array operations here and in the next few lines.

mibidata/constants.py
```diff
@@ -0,0 +1,15 @@
+"""Constants used in mibidata module.
+
+Copyright (C) 2020 Ionpath, Inc.  All rights reserved."""
+
+COLORS = {
```
jaytarolli: I like that these are defined in RGB instead of the hue. It's easier for me to recogize.

mibidata/color.py
rfinck: That was intentional, using the same array as input and output in `ndimage.filters.convolve` changes the data as it's processing and results in a different output than desired.

PR #88, opened 2020-03-02
DPPA-714 Update overlay code to be compatible with MIBItracker v1.1.8

rfinck:
We expect to release MIBItracker v1.1.8 next week, which has some changes to how the overlay sliders work and therefore what the downloadable json settings are. This should be released at the same time to maintain compatibility.


mibidata/color.py
```diff
@@ -252,28 +245,33 @@ def compose_overlay(image, overlay_settings):
...
-        if item.get('blur', 0) > 0:
-            ndimage.filters.gaussian_filter(
-                array, item['blur'] * 100, output=array)
+        array = ndimage.filters.convolve(
+            array, constants.OVERLAY_SMOOTHING_KERNELS[setting['blur']])
         np.clip(array, 0, 1, out=array)
+        if setting['brightness'] > 0:
+            array /= (1 - setting['brightness'])
+            np.clip(array, 0, 1, out=array)
+        elif setting['brightness'] < 0:
```
yari-ionpath: how come brightness has different effects for positive (linear) and negative (non-linear) values?

mibidata/constants.py
```diff
@@ -13,3 +13,91 @@
     'Blue': (0.0, 0.0, 1.0),
     'Gray': (1.0, 1.0, 1.0),
 }
+
+OVERLAY_MIN_BRIGHTNESS = -0.9
+
+# The kernels below are those used by WegGL convolution in MIBItracker.
+# pylint: disable=line-too-long
+OVERLAY_SMOOTHING_KERNELS = [
+    [
```
yari-ionpath: the first array here (a blur value of 0, seeming no blur) appears to be out of order. If this is the desired kernel, it would be good to add some explanation as to why.

mibidata/color.py
yari-ionpath: also this math would seem to support the fact that a max_range=10 and brightness 0 should be equivalent to max_range=1 and brightness 0.9, correct? I would not expect similar equivalency if brightness<0.

PR #90, opened 2020-04-17
DPPA-869 Add refresh route and default to automatically call it every 5 minutes.

rfinck:
This enables a token to be refreshed so it stays valid during a long-running (hours) script, and becomes more necessary as default token expiration times get shorter. They are anticipated to become 45 minutes in the near future.


rfinck:
@yari-ionpath pillow is a dependency of skimage, and under the hood the version that was being installed changed, resulting in tests failing that also indicate this would be a failure in normal usage. You are right that it's not clear why it's part of this PR (this just happens to be the first PR since that change), so I will follow better practices than I did here and put up a separate PR that if it's merged first can prevent this one from failing.

rfinck:
@yari-ionpath This needs another review now that I reverted those unrelated commits.

yari-ionpath:
great. thanks for the explanation (and setting a good example for task specific PRs!)

rfinck:
@jaytarolli I am going to merge this, but will wait to tag a release until you've confirmed you don't need additional changes for your dependencies.

PR #91, opened 2020-04-17
DPPA-870 Specify pillow version

rfinck:
This prevents imread from failing when using the PIL plugin.


jaytarolli:
@rfinck Is there a reason `pillow==4.3.0` was used? Creating a conda environment with that version doesn't work on Windows, though will on a Linux VM. I am able to create the environment if I specify 6.2.1, which is the version that is currently on my VM which worked fine reading PNG files last weekend (as opposed to 7.x which was installed on my PC).

From the [installation page](https://pillow.readthedocs.io/en/latest/installation.html#windows-installation), pillow versions for Windows are packaged with zlib for supported versions of Python. The version you use here does not support Python 3.7, so when installing 4.3.0, it fails because there isn't already a precompiled version and attempts to compile with zlib, which is not installed on my computer. The version of pillow needs to be >=5.2 for Python 3.7.

rfinck:
@jaytarolli Probably because I looked at one machine that had mibitracker-client installed and checked what version it was using, rather than taking the thorough look that you did. Happy to bump it to 6.2.1.

PR #94, opened 2020-04-23
DPPA-870-fix Use the latest compatible version of pillow

rfinck:
Upping the pillow version to the latest that works, not a super old one that happened to be lying around.

I was hoping (without lyet ooking into the details) that this would be fixed in the latest pillow release, and will keep an eye out for when we can revert this and let skimage choose for us. https://github.com/python-pillow/Pillow/blob/master/CHANGES.rst#711-2020-04-02


rfinck:
@jaytarolli If this looks ok I can add a commit to bump the version in setup.py and then make a release.

PR #95, opened 2020-05-06
TKR-688 Include run_id when requesting signed url for uploading tiff

rfinck:
An upcoming release of MIBItracker will require the run id to be specified when initiating a tiff upload. Adding this parameter now allows it to continue to work with the current release (V1.2.0) and then still work once this is enforced in a later version.


rfinck:
@rileysparsons This is the companion PR to the larger one you just reviewed, and needs to go in first.

PR #99, opened 2020-08-26
DPPA-1123 Rename mibitracker-client to mibilib

rfinck:


environment.yml
```diff
@@ -1,11 +1,12 @@
-name: mibitracker-client
+name: mibilib
 channels:
 - conda-forge
 - defaults
 dependencies:
 - pip=19.1.1
 - python=3.7.3
 - pip:
+    - astroid==2.2.5
```
murataksoy, line 9: uuuhhhh... what's this? And why does it conflict with the mibitof android version?

murataksoy: You know how I feel about environment changes :-)

murataksoy, line 9: Why was this change necessary? It seems to conflict with the mibitof version.

rfinck: astroid is a dependency of pylint and the version it was pulling caused it to error (see circleci failure). ultimately we should fix it by updating the pylint version but that should be a separate ticket since we may need to adjust settings. https://github.com/PyCQA/pylint/issues/2388

PR #102, opened 2020-09-18
DPPA-436: Document loading and segmenting mibitiff in cell profiler and Qupath

jaytarolli:


scripts/qupath_export_label_image.groovy
```diff
@@ -0,0 +1,45 @@
...
+    }
+}
+
+g2d.dispose()
+
+new ImagePlus('Labels', labelImage).show()
+
+def name = getProjectEntry().getImageName()
+def path = buildFilePath(PROJECT_BASE_DIR)
+def fileoutput = new File(path, 'labels.png')
```
rfinck, line 44: This saves the file as 'labels.png' in the top level of the project directory, but if I have multiple images in my project, then it'll overwrite each one if I run it on each.

scripts/qupath_export_label_image.groovy
```diff
@@ -0,0 +1,45 @@
...
+ImageIO.write(labelImage, 'png', fileoutput)
```
rfinck, line 45: Please add a newline at the end.

rfinck (commented):
I would suggest putting this in a `qupath` or `qupath_scripts` or `scripts/qupath` folder rather than simply `scripts`.

scripts/qupath_export_label_image.groovy
jaytarolli: I can append "labels" to the end of the name as supplied by QuPath, but it is formatted "[original file name].tiff - [original file name].tiff SIMS - labels.png". That seems to be a limitation of how QuPath builds an image name and I don't immediately see a way around it.

rfinck: I was going to suggest some string manipulation such as 
```
def imageName = getProjectEntry().getImageName()
def originalName = imageName.split('.tiff - ')[0]
```
but I believe the `getOriginalImageName` method is actually what you want:
```
def imageName = getProjectEntry().getOriginalImageName()
```

rfinck:
@jaytarolli What's the state of this? Is it awaiting another review?

jaytarolli:
@rfinck I believe I was waiting on your input regarding my last comment about the output file name.

rfinck (approved):
Sorry for the super long delay, feel free to remind me next time!

PR #109, opened 2021-03-30
DPPA-1411 Add set_fov_id to change both fov_id and folder at same time

rfinck:


mibidata/tests/test_mibi_image.py
```diff
@@ -138,6 +138,23 @@ def test_check_folder_without_fov_id(self):
...
+
+    def test_set_fov_id_bad_folder(self):
+        image = mi.MibiImage(TEST_DATA, TUPLE_LABELS, fov_id='FOV1')
+        with self.assertRaises(ValueError):
+            image.set_fov_id('FOV2', 'FOV3')
+
+    def test_set_fov_id_with_folder(self):
+        image = mi.MibiImage(TEST_DATA, TUPLE_LABELS,
+                             folder='Point1/RowNumber0/Depth_Profile0')
+        image.set_fov_id('Point2', 'Point2/RowNumber0/Depth_Profile0')
```
murataksoy, line 155: You can add one more test here where you do `image.set_fov_id('Point1')` without the `folder`argument, and check if folder stays at `'Point1/RowNumber0/Depth_Profile0'`

murataksoy (approved):
A minor comment, otherwise this is what we needed.

mibidata/tests/test_mibi_image.py
rfinck: I claim that is the same as the first test.

murataksoy (approved):

PR #110, opened 2021-03-30
DPPA-1444 Add a pixel_size property to MibiImage

rfinck:


mibidata/tests/test_mibi_image.py
```diff
@@ -90,6 +90,16 @@ def test_obj_datetime(self):
         image = mi.MibiImage(TEST_DATA, TUPLE_LABELS, date=date)
         self.assertEqual(image.date, date)
 
+    def test_pixel_size(self):
+        size = 500
+        image = mi.MibiImage(TEST_DATA, TUPLE_LABELS, size=size)
+        self.assertEqual(image.pixel_size, size / 2)  # test data is 2 x 2 x 3
+
+    def test_no_pixel_size(self):
+        image = mi.MibiImage(TEST_DATA, TUPLE_LABELS)
```
murataksoy, line 99: I didn't know you could crate a MibiImage without the size.

murataksoy (approved):

mibidata/tests/test_mibi_image.py
rfinck: You need the size to save a tiff, but the class instance doesn't require it. This is a holdover from the early alpha days.

PR #111, opened 2021-04-10
DPPA-1471 Bump skimage and related packages

rfinck:
The following packages were updated to the listed versions:
matplotlib==3.3.2
numpy==1.20.0
pandas==1.2.3
pillow==8.2.0
scikit-image==0.18.0
scikit-learn==0.23.2


stanislaw-ionpath (approved):
I was really close... I only missed 🐍 -> 🐪 transition. 

Good job!

stanislaw-ionpath:
Rachel, 
1. What needs to happen to have it merged to master?
2. Maybe we could bump other libraries too?

setup.py
```diff
@@ -13,13 +13,14 @@
       license='GNU General Public License v3.0',
       python_requires='~=3.6',
       install_requires=[
```
stanislaw-ionpath, line 15: One thing that is more for my education. 
Why the list of packages in `install_requires` is shorter than `- pip:` section in `environment.yml`?

MibiImage_Tutorial.ipynb
```diff
@@ -2003,7 +2003,7 @@
     "    \"\"\"Scales each 2-D slice of an N-D image to a brightened version in the [0, 1] range.\"\"\"\n",
     "    \n",
     "    scaled = np.divide(\n",
-    "        image.data.astype(np.float),\n",
+    "        image.data.astype(float),\n",
```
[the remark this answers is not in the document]
rfinck: Hint: Look at the commit message.

https://numpy.org/devdocs/release/1.20.0-notes.html#using-the-aliases-of-builtin-types-like-np-int-is-deprecated

setup.py
rfinck: The list in `install_requires` is sufficient for _installing_ mibilib.

The list in `environment.yml` is for _development_. It therefore also includes packages for testing, generating documentation, etc.

These two options are noted in the readme as Option A and Option B: https://github.com/ionpath/mibilib#option-a-development-clone-repository-and-create-environment

rfinck: I just noted that `setup.py` uses an older version of Python, we should bump that here to get it to match the version in `environment.yml`.

rfinck:
>What needs to happen to have it merged to master?
Let's confirm that it works for your ultimate use-case, which is outside of this package but can be tested with the `tests/DPPA-1471` tag. Please update the Python version in `setup.py` as I noted above as well.

>Maybe we could bump other libraries too?
I'd vote that we don't increase the scope of this PR.

setup.py
stanislaw-ionpath: Done

stanislaw-ionpath:
@rfinck when shall we bump the mibilib version in setup.py?

rfinck:
>when shall we bump the mibilib version in setup.py?

Now :)

PR #112, opened 2021-04-20
DPPA-1382 Tiling

murataksoy:
The tiling script here will be used to generate a json file containing a tile of FOVs. The input will be a single FOV list json exported from instrument control software. The generated tiled json file can be imported back to the Instrument Control Software.

### DIFF
--- a/scripts/tests/test_tiling.py
+++ b/scripts/tests/test_tiling.py
@@ -1,0 +1,710 @@
+import os
+import unittest
+import tempfile
+import json
+import numpy as np
+from tiling import tiling
+
+FOV_LIST_JSON_ORIG = {
+    "exportDateTime": "2020-09-04T15:13:58",
+    "fovFormatVersion": "1.5",
+    "fovs": [
+        {
+            "scanCount": 1,
+            "centerPointMicrons": {
+                "x": 123000.0,
+                "y": 45000.0
+            },
+            "fovSizeMicrons": 800,
+            "timingChoice": 1,
+            "frameSizePixels": {
+                "width": 64,
+                "height": 64
+            },
+            "imagingPreset": {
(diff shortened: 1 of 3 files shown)


scripts/tiling/tiling.py
```diff
@@ -0,0 +1,72 @@
...
+
+def tile(fov_list_json_file, xn, yn, overlap):
+    ''' Using a template json file, creates another fov json that includes
+    the tiled version of the original FOV.
+    Args:
+        fov_list_json_file: The FOV json containing the FOV to be tiled. The
+            resulting tiled json is created in the same directory as this file.
+        xn: The number of tiles in the x direction.
+        yn: The number of tiles in the y direction.
+        overlap: The degree of overlap between tiles. Must be between 0-1.
```
jaytarolli, line 21: A use case for some could be to leave spacing between adjacent FOVs instead of overlapping. Additionally, it would be useful to be able to specify different overlap/spacing in each the x and y dimensions.

scripts/tests/test_tiling.py
```diff
@@ -0,0 +1,714 @@
+''' Unit tests for tiling module. '''
+
+import os
```
jaytarolli, line 3: It doesn't appear these tests got run in CircleCI. Do you need to add the folder to the config?

image

jaytarolli (changes requested):

scripts/tiling/tiling.py
murataksoy: Done

scripts/tests/test_tiling.py
murataksoy: Done

jaytarolli (approved):

PR #115, opened 2021-07-13
DPPA-1539 Unable to generate MIBItiff if any of the channels contain gamma in its name

murataksoy:
Happens only in VMs used for research-services.

Use conda list to compare instrument vs VM library versions.

For reproducing the error, use the latest mibilib and call the following using the tiff file below:

tiff.write('/Users/murat/Downloads/test.tiff', tiff.read('/Users/murat/Downloads/20210613_6247_Run-3864_FOV2_ROI1.tiff'))

Using utf-8 instead of ascii in tifffile.py L1902 works, but this is not a good solution since it requires a custom change to the underlying library.

The solution proposed here converts the page tag into bytes string so that tifffile.py does not try to convert it to ascii.

### DIFF
--- a/mibidata/tiff.py
+++ b/mibidata/tiff.py
@@ -165,7 +165,7 @@ def write(filename, image, sed=None, optical=None, ranges=None,
                     'channel.target': image.targets[i],
                 })
                 page_name = (
-                    285, 's', 0, '{} ({})'.format(image.targets[i],
+                    285, 's', 0, '{} ({})'.format(image.targets[i].encode(),
                                                   image.masses[i])
                 )
                 min_value = (340, range_dtype, 1, ranges[i][0])
--- a/mibidata/tiff.py
+++ b/mibidata/tiff.py
@@ -212,7 +212,7 @@ def write(filename, image, sed=None, optical=None, ranges=None,
                 'channel.target': image.targets[i],
             })
             page_name = (285, 's', 0, '{} ({})'.format(
-                image.targets[i], image.masses[i]))
+                image.targets[i].encode(), image.masses[i]))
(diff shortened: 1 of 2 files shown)


mibidata/tiff.py
```diff
@@ -212,7 +212,7 @@ def write(filename, image, sed=None, optical=None, ranges=None,
                 'channel.target': image.targets[i],
             })
             page_name = (285, 's', 0, '{} ({})'.format(
-                image.targets[i], image.masses[i]))
```
stanislaw-ionpath, line 215: What encode will do here with γ, β, μ, π? will it replace it with \\u03b2 \\u03b3 \\u03b4 \\u03c0? IUs taht what we want?

stanislaw-ionpath (commented):
Murat, i do not fully understand what is going on here. Is testing in CircleCI enough to proove that this is working?

mibidata/tiff.py
murataksoy: What it does is that it converts it to bytes string, and when writing to mibitiff, it gets converted back to utf-8. 

If you enter these directly as strings, tifffile tries to convert them to ascii, and things catch on fire. 

I'm not 100% sure why it is trying to convert strings to ascii but does not do the same with byte strings, but it works somehow.

I will ask someone else from the instrument control team to look at this as well.

stanislaw-ionpath: @murataksoy thanks for explanation I think it makes sense.  

@rileysparsons I hold on with my approval waiting for your comments. You can approve yourself if you think it is fine.

rileysparsons (approved):
Fixes the problem and makes sense.

PR #116, opened 2021-07-15
DPI-9 Increment mibilib version in setup.py to match the next release

murataksoy:


stanislaw-ionpath (commented):
OK, so you bump the version here, commit it and then make a tag with the same version number. correct?

murataksoy:
> OK, so you bump the version here, commit it and then make a tag with the same version number. correct?

Correct.

stanislaw-ionpath (approved):

PR #124, opened 2022-10-26
Changing requests to minimum version

SpacePirate:
Changing to  `requests>=2.20.1` in `environment.yml` file to allow dependency compatibility with icb-client repo


therealjacinto (approved):
🌮

PR #127, opened 2022-11-23
task/MIB-19_upgrade_python_3_11

avstri:


therealjacinto (approved):
tested and looks good (but might want @jaytarolli's approval). Tested with installing locally and installing other ICB packages with the commit hash

PR #131, opened 2024-05-01
Mib 21: Add stitching script

mlnagy:


environment.yml
```diff
@@ -13,7 +13,7 @@ dependencies:
   - matplotlib==3.6.2
   - mock==4.0.3
   - numpy==1.23.5
-  - pandas==1.2.3
```
jaytarolli: Are you sure this doesn't break anything? What version do you need for this script to work?

scripts/tiling/stitching.py
```diff
@@ -0,0 +1,256 @@
...
+                        time.sleep(0.50)
+                    else:
+                        mr.upload_mibitiff(out_path, run_id=ref_json['run']['id'])
+            
+            print(f"Stitched MIBItiff, {out_path} uploaded to MIBItracker.")
+                        
+
+if __name__ == "__main__":
+
+    fovs_folder = "/Users/mnagy/projects/stitch_script_test/2024-04-29T10-07-46_gold_tonsil_3x3_coarse_ROI"
```
jaytarolli: You should remove paths to your local machine. Generally, if there are variables that a user needs to set in order for the script to work, it should be indicated and some documentation included. For example: `fovs_folder = '' # Set this to the path of the folder that contains the FOVs to stich`

scripts/tiling/stitching.py
```diff
@@ -0,0 +1,256 @@
...
+    out_path = "/Users/mnagy/projects/stitch_script_test/2024-04-29T10-07-46_gold_tonsil_3x3_coarse_ROI.tiff"
+    cmd = f'ls {fovs_folder}/*.tiff'
```
jaytarolli: Scripts needs to be compatible across platforms (macOS, Linux, and Windows), and I don't think `ls` will always work on a Windows machine. Instead, anytime you're interacting with paths in the filesystem, I would use standard Python libraries as they will handle dealing with the platform. So for here, you could use `os.listdir()`. If you need to select files that match a given pattern, you can `import glob`.

scripts/tiling/stitching.py
```diff
@@ -0,0 +1,256 @@
...
+        fov_paths.remove("")
+    except:
+        pass
+
+    session_dict = {
+        "url": "https://mibitracker.api.ionpath.com/",  # MIBItracker backend URL
+        "email": "",  # User name
+        "password": ""  # User password
+    }
+    mt_upload = True
```
jaytarolli: For a non Ionpath user `mt` isn't really descriptive, consider changing to `mibitracker_upload` or `do_mibitracker_upload`.

scripts/tiling/stitching.py
```diff
@@ -0,0 +1,256 @@
...
+
+        cols = []
+        rows = []
+        for fov_path in roi_fov_paths:
+            fov_name = os.path.basename(fov_path).split("-")[-1].split(".tiff")[0]
+            c, r = fov_name.split("_")[1:]
+            cols.append(int(c[1:].lstrip("0")))
+            rows.append(int(r[1:].lstrip("0")))
+
+        out_img, panel, max_dim = combine_entity_by_name(roi_fov_paths, cols, rows, enforce_square=True)
```
jaytarolli: MIBIcontrol allows users to specify a positive or negative spacing between tiles, but this script seems to assume no padding or overlap between adjacent tiles. I think that should be included when reconstructing since it's a parameter that's available to set and the value (can be specified for each the x- and y-dimension) is available in the run JSON file.

scripts/tiling/stitching.py
```diff
@@ -0,0 +1,256 @@
...
+                    ref_json = mr.get('images/{}/'.format(ref_image_id)).json()
+
+        sample_id = run+"__"+roi
+        f_split = ref_json['folder'].split('/')
+        f_split[0] = sample_id
+        folder_name = '/'.join(f_split[:3])
+
+        px_per_u = ref_json["frame"]/ref_json["fov_size"]
+
+        metadata = {
```
jaytarolli: Is this really needed? The majority of metadata should be the same across all FOVs within a ROI (except coordinates, obviously). And in that case, you could just copy the metadata from the first TIFF file you read in.

scripts/tiling/stitching.py
```diff
@@ -0,0 +1,256 @@
...
+
+        img_shape = list(np.shape(fov_img))
+        ch_count = np.min(img_shape)
+        shape_2d = img_shape[:-1]
+
+        if out_img is None:
+            out_img = np.zeros((h*shape_2d[0], w*shape_2d[1], ch_count), dtype=fov_img.dtype)
+            out_img_shape = list(np.shape(out_img))
+
+        for ch_i in range(ch_count):
```
jaytarolli: Why loop through all the channels instead of using something like `out_image[startY:endY, startX:endX, :] = fov_img[:, :, :]`?

scripts/tiling/stitching.py
```diff
@@ -0,0 +1,256 @@
...
+            out_img[((rows[fov_i]-1)*shape_2d[0]):(rows[fov_i]*shape_2d[0]), 
+                    ((cols[fov_i]-1)*shape_2d[1]):(cols[fov_i]*shape_2d[1]), ch_i] = fov_img[:,:,ch_i]
+    
+    if enforce_square:
```
jaytarolli: Is `enforce_square` needed after @therealjacinto merges in changes that enable non-square TIFF writing?

jaytarolli (changes requested):

environment.yml
mlnagy: Oops! That was me doing some local tests earlier this year and it slipped in. Adding versioning back in.

scripts/tiling/stitching.py
mlnagy: It's an artifact of our internal script where this functions for situations where the files are separate and needed to be loaded and added separately. Updating to your recommendation.

scripts/tiling/stitching.py
mlnagy: Adding this in would take me some time, as I have never looked at that parameter (nor was aware of it). The situation where we add spacing is more straightforward, but a negative spacing would need some decisions to be made regarding how it's all handled: Do we blend them all, have the first FOV's value take precedence, or the last?

scripts/tiling/stitching.py
mlnagy: Hmm, yeah, good point. That will mean we won't need to connect to MIBItracker at all if we are not uploading to it.

scripts/tiling/stitching.py
jaytarolli: I think we do need to include it. That said, the intent of this script is to be a usable tool for users but more importantly to be a starting point reference for users who want/need to do something more complex. So, we don't necessarily need to solve the problem (specifically for overlapping tiles) in a robust manner. A simple solution may be all that's needed here.

For the negative spacing, it can be as simple as pixels being overwritten as the image is stitched together. If you iterate through the y tiles and then the x tiles from 0 to n, then the pixels will be overwritten in a predictable way. And that should suffice for the purpose of a working example.

For the actual spacing/overlap parameter, you could take one of two approaches.
1. Load the run JSON file that's in the folder with all the TIFFs and use the `xMargin` and `yMargin` values. The downside here is that MIBIcontrol defines them in microns, but you need pixels. So, you'd need to read the metadata of one of the MIBItiffs to get the pixel scaling of the images and convert microns to pixels.
2. Add xMargin and yMargin (in pixel units) as arguments to the script or as variables that can be set. They should default to 0, and then a user can override them.

mlnagy: Thanks for these, it was very helpful! Will also incorporate this into our workflow. I added this in the latest commit (both 1 and 2 functionality is there).

scripts/stitching/stitching.py
```diff
@@ -1,24 +1,27 @@
...
-from mibitracker.request_helpers import MibiRequests
+import sys
 import time
+
+import numpy as np
+
+from mibidata import mibi_image as mi
 from mibidata import tiff
-import json
+from mibitracker.request_helpers import MibiRequests
```
jaytarolli, line 16: Reordered the imports. Python style is to import 1) standard modules 2) modules installed with conda/pip 3) custom modules, such as those in the repo being worked on

scripts/stitching/stitching.py
```diff
@@ -32,203 +35,291 @@ def combine_entity_by_name(roi_fov_paths, cols, rows, enforce_square=False, marg
 
         img_shape = list(np.shape(fov_img))
         ch_count = np.min(img_shape)
-        shape_2d = [int(v) for v in img_shape[:-1]]
+        shape_2d = img_shape[:-1]
 
         if out_img is None:
-            out_img = np.zeros((h*shape_2d[0]+(h-1)*margin["y"], w*shape_2d[1]+(w-1)*margin["x"], ch_count), 
-                               dtype=fov_img.dtype)
+            out_img = np.zeros(
```
jaytarolli, line 43: Here, and a bunch of other places, kept line lengths to <=80 characters

scripts/stitching/stitching.py
```diff
@@ -32,203 +35,291 @@ def combine_entity_by_name(roi_fov_paths, cols, rows, enforce_square=False, marg
...
+                um_min_x = coord["x"]
+            if coord["y"] < um_min_y:
+                um_min_y = coord["y"]
+            roi_fov_paths.append(
+                os.path.join(
+                    run_folder,
+                    "fov" + "-" + str(
+                        f["runOrder"]).zfill(2) + "-" + fov_name) + ".tiff")
+    
+        print(f'Stitching {run} - {roi}: {len(roi_fov_paths)} FOVs')
```
jaytarolli, line 107: Made the print statement a bit more informative/readable

scripts/stitching/stitching.py
```diff
@@ -32,203 +35,291 @@ def combine_entity_by_name(roi_fov_paths, cols, rows, enforce_square=False, marg
...
-                out_mibi_tiff.set_image_size(dims[0]/px_per_u)
-            
-            f_split = out_mibi_tiff.folder.split('/')
-            f_split[0] = roi
-            out_mibi_tiff.set_fov_id(f_split[0], '/'.join(f_split))
+        f_split = out_mibi_tiff.folder.split('/')
+        f_split[0] = roi
+        out_mibi_tiff.set_fov_id(f_split[0], '/'.join(f_split))
 
+        if not out_folder:
```
jaytarolli, line 125: Made `out_folder` an optional parameter, if omitted then it defaults to saving in the `run_folder`

scripts/stitching/stitching.py
```diff
@@ -32,203 +35,291 @@ def combine_entity_by_name(roi_fov_paths, cols, rows, enforce_square=False, marg
...
-                fov_name = f["name"]
-                cols.append(f["gridPosition"]["x"]+1)
-                rows.append(f["gridPosition"]["y"]+1)
-                coord = f["centerPointMicrons"]
-                if coord["x"] < um_min_x:
-                    um_min_x = coord["x"]
-                if coord["y"] < um_min_y:
-                    um_min_y = coord["y"]
-                roi_fov_paths.append(os.path.join(run_folder,"fov"+"-"+str(f["runOrder"]).zfill(2)+"-"+fov_name)+".tiff")
+    for r in run_json["rois"]:
```
jaytarolli, line 73: I pulled out all the rest of the code from the `with` block

scripts/stitching/stitching.py
```diff
@@ -32,203 +35,291 @@ def combine_entity_by_name(roi_fov_paths, cols, rows, enforce_square=False, marg
...
+                ['fov']['section']['id']
+            new_im_metadata['aperture'] = metadata['aperture']
+            new_im_metadata['imaging_preset'] = metadata['imaging_preset']
+            new_im_metadata['lens1_voltage'] = metadata['lens1_voltage']
+
+            if not full_id:
                 for t in range(MAX_TRIES):
                     try:
-                        run_id = mr.get('/runs/', params={'name': run}).json()['results'][0]['id']
+                        _ = mr.post('/images/', json=new_im_metadata)
```
jaytarolli: No need for the ignored return value

scripts/stitching/stitching.py
```diff
@@ -32,203 +35,291 @@ def combine_entity_by_name(roi_fov_paths, cols, rows, enforce_square=False, marg
...
+                except:
+                    if t < MAX_TRIES-1:
+                        time.sleep(0.50)
+                    else:
+                        mr.upload_mibitiff(out_path, run_id=run_id)
+            
+            print(
+                f"Stitched MIBItiff, {out_path}, uploaded to MIBItracker.")
+
+def parse_args(args):
```
jaytarolli, line 236: Instead of specifying variables in the script itself, let the users pass them as arguments when running it

scripts/stitching/stitching.py
```diff
@@ -32,203 +35,291 @@ def combine_entity_by_name(roi_fov_paths, cols, rows, enforce_square=False, marg
...
+    ''' Argument parsing helper function.   
+    '''
+    parser = argparse.ArgumentParser(
+        description='Script for creating a stitched ROI MIBItiff from a folder '
```
jaytarolli, line 240: Moves the script description and argument help text to argparse so that it can be viewed with `python stitching.py --help`

jaytarolli:
@mlnagy I made some changes and have added comments to my changes to give you context for why I did. I also tested it to some degree. Testing performed:
- `python .\stitching.py --run_folder "C:/Users/Jay Tarolli/data/M14/2024-05-30T21-45-35-7pass_largeROI"`: Success
- `python .\stitching.py --run_folder "C:/Users/Jay Tarolli/data/M14/2024-05-30T21-45-35-7pass_largeROI" --out_folder 'C:/Users/Jay Tarolli/data/M14/2024-05-30T21-45-35-7pass_largeROI_subset/subfolder'`: Success
- `python .\stitching.py --run_folder "C:/Users/Jay Tarolli/data/M14/2024-05-30T21-45-35-7pass_largeROI" --enforce_square`: Sets `enforce_square` to True; canceled before TIFF generation
- `python .\stitching.py --run_folder "C:/Users/Jay Tarolli/data/M14/2024-05-30T21-45-35-7pass_largeROI" --upload_to_mibitracker`: Sets `upload_to_mibitracker` to True, fails because URL, email, and password not specified
- `python .\stitching.py --run_folder "C:/Users/Jay Tarolli/data/M14/2024-05-30T21-45-35-7pass_largeROI" --upload_to_mibitracker --mibitracker_email [redacted] --mibitracker_password [redacted] --mibitracker_url [redacted]`: MibiRequests instance created, verifying log in; canceled before upload
- `python .\stitching.py --run_folder "C:/Users/Jay Tarolli/data/M14/2024-05-30T21-45-35-7pass_largeROI"`: verified `margin = {'x': 0, 'y': 0}`
- `python .\stitching.py --run_folder "C:/Users/Jay Tarolli/data/M14/2024-05-30T21-45-35-7pass_largeROI" --fov_margin_size_x 10`: verified `margin = {'x': 10, 'y': 0}`
- `python .\stitching.py --run_folder "C:/Users/Jay Tarolli/data/M14/2024-05-30T21-45-35-7pass_largeROI" --fov_margin_size_x 10 --fov_margin_size_y 2`: verified `margin = {'x': 10, 'y': 20}`

scripts/stitching/stitching.py
```diff
@@ -0,0 +1,325 @@
...
+
+import numpy as np
+
+from mibidata import mibi_image as mi
+from mibidata import tiff
+from mibitracker.request_helpers import MibiRequests
+
+MAX_TRIES = 5
+
+def combine_entity_by_name(roi_fov_paths, cols, rows, enforce_square, margin):
```
therealjacinto, line 20: I see that we probably won't be needing `enforce_square` anymore now that mibitracker allows for non-square images. Do we want to at least change this so that it has the default value of `False`?

scripts/stitching/stitching.py
```diff
@@ -0,0 +1,325 @@
...
+                out_img,((0,0),(0,out_img_shape[0]-out_img_shape[1]),(0,0)))
+        elif out_img_shape[0] < out_img_shape[1]:
+            out_img = np.pad(
+                out_img,((0,out_img_shape[1]-out_img_shape[0]),(0,0),(0,0)))
+        
+    return out_img.astype(out_img.dtype, copy=False), panel, \
+        int(max(w*shape_2d[1], h*shape_2d[0])), ref_metadata
+
+
+def stitch_fovs(run_folder, out_folder, session_dict, upload_to_mibitracker,
```
therealjacinto, line 67: to make it clear that `out_folder` can be `None`, can we have it have the default value of `None`?

scripts/stitching/stitching.py
therealjacinto: See comment

scripts/stitching/stitching.py
```diff
@@ -0,0 +1,325 @@
...
+import sys
+import time
+
+import numpy as np
+
+from mibidata import mibi_image as mi
+from mibidata import tiff
+from mibitracker.request_helpers import MibiRequests
+
+MAX_TRIES = 5
```
therealjacinto, line 18: Can we rename this and have a comment on what this is used for? Something like `MAX_MIBITRACKER_REST_TRIES` or something

scripts/stitching/stitching.py
```diff
@@ -0,0 +1,325 @@
+""" Script for creating a stitched ROI MIBItiff from a folder of FOVs.
```
therealjacinto, line 1: All methods in this file need a docstring explaining what the function does and what each argument is: https://peps.python.org/pep-0257/#multi-line-docstrings

therealjacinto (changes requested):

scripts/stitching/stitching.py
```diff
@@ -0,0 +1,327 @@
...
+
+import argparse
+import json
+import os
+import sys
+import time
+
+import numpy as np
+
+sys.path.append('../..')
```
jaytarolli, line 12: I think this is needed if a user is just using the mibilib conda env

PR #132, opened 2024-07-19
Mic 3034

DeeeAich:
Created support prepared for python 3.12, made some code changes to fix the new API changes of new versions


DeeeAich:
I've noticed the check has failed, I'm working through fixing this problem, it seems that py and potentially py.test is not working/ is missing

setup.py
```diff
@@ -11,17 +11,17 @@
       url='https://github.com/ionpath/mibilib',
       description='Python utilities for IONpath MIBItracker and MIBItiff data',
       license='GNU General Public License v3.0',
-      python_requires='~=3.11.0',
+      python_requires='>=3.9',
```
therealjacinto: Per the ticket, this needs to support at least `3.8`

therealjacinto (changes requested):
See comment

